PR #10, opened 2019-10-24
Speed up load action

nikita-klsh:
This PR 

1.  Speed up `load` action for `segy` files via ignoring cube geometry inference when opening `segy`.
2. Avoid unnecessary usage of list comprehensions in some actions.


seismicpro/src/seismic_batch.py
```diff
@@ -1121,7 +1123,7 @@ def mask_to_pick(self, src, dst, labels=True):
             data = np.argmax(data, axis=1)
 
         dst_data = massive_block(data)
-        setattr(self, dst, np.array([i for i in dst_data] + [None])[:-1])
+        setattr(self, dst, np.array(dst_data + [None])[:-1])
```
roman-kh: Leave a comment that an array is implicitly converted to object dtype.

dpodvyaznikov:
@a-arefina at this moment actions do not work properly, so adding them will be of no use. I expect to fix it soon

PR #12, opened 2019-11-15
Update `drop_zero_traces`

dpodvyaznikov:
And fix `sort_traces` action.

### DIFF
--- a/seismicpro/src/seismic_batch.py
+++ b/seismicpro/src/seismic_batch.py
@@ -665,7 +665,7 @@ class SeismicBatch(Batch):
             Batch with new trace sorting.
         """
         _ = args
-        sorting = self.meta[dst]['sorting']
+        sorting = self.meta[src]['sorting']
 
         pos = self.get_pos(None, src, index)
         df = self.index.get_df([index])
@@ -678,7 +678,7 @@ class SeismicBatch(Batch):
             sorted_index_df = df[cols].sort_values(sorting)
             order = np.argsort(sorted_index_df[sort_by].values)
         else:
-            order = np.argsort(df[sort_by].tolist())
+            order = np.argsort(df[sort_by].values)
 
         getattr(self, dst)[pos] = getattr(self, src)[pos][order]
 
@@ -741,10 +741,12 @@ class SeismicBatch(Batch):
         for comp in self.components:
             getattr(self, comp)[pos] = getattr(self, comp)[pos][mask]
 
-        cols = [(INDEX_UID, src), (sorting, '')]
-        sorted_index_df = self.index.get_df(index)[cols].sort_values(sorting)
-        order = np.argsort(sorted_index_df[cols[0]].values)
-        return mask[order]
+        if sorting:
+            cols = [(INDEX_UID, src), (sorting, '')]
+            sorted_index_df = self.index.get_df([index])[cols].sort_values(sorting)
+            order = np.argsort(sorted_index_df[cols[0]].values)
+            return mask[order]
+        return mask
 
     @action
     @inbatch_parallel(init='_init_component')
--- a/seismicpro/src/seismic_batch.py
+++ b/seismicpro/src/seismic_batch.py
@@ -172,7 +172,7 @@ class SeismicBatch(Batch):
 
     def _post_filter_by_mask(self, mask, *args, **kwargs):
         """Index filtration using all received masks. This post function assumes that
-        components have already been sorted.
+        components have already been filtered.
 
         Parameters
         ----------
@@ -188,7 +188,8 @@ class SeismicBatch(Batch):
         ----
         Batch items in each component should be filtered in decorated action.
         This post function creates new instance of SeismicBatch with new index
-        instance.
+        instance and copies filtered components from original batch for elements
+        in new index.
         """
         _ = args, kwargs
         if any_action_failed(mask):
--- a/seismicpro/src/seismic_batch.py
+++ b/seismicpro/src/seismic_batch.py
@@ -9,7 +9,7 @@ import segyio
 
 from ..batchflow import action, inbatch_parallel, Batch, any_action_failed
 
-from .seismic_index import SegyFilesIndex, FieldIndex
+from .seismic_index import SegyFilesIndex, FieldIndex, TraceIndex
 
 from .utils import (FILE_DEPENDEND_COLUMNS, partialmethod, calculate_sdc_for_field, massive_block,
                     check_unique_fieldrecord_across_surveys)
@@ -205,17 +205,16 @@ class SeismicBatch(Batch):
                                                   index_name=self.index.name)
 
         new_batch = type(self)(batch_index)
-        new_batch.add_components(self.components)
+        new_batch.add_components(self.components, len(self.components) * [new_batch.array_of_nones])
         new_batch.meta = self.meta
 
-        for comp in new_batch.components:
-            setattr(new_batch, comp, np.array([None] * len(new_batch.index)))
-
-        for index in new_index:
-            for isrc in new_batch.components:
-                pos_new = new_batch.get_pos(None, isrc, index)
-                pos_old = self.get_pos(None, isrc, index)
-                getattr(new_batch, isrc)[pos_new] = getattr(self, isrc)[pos_old]
+        for isrc in new_batch.components:
+            pos_new = []
+            pos_old = []
+            for index in new_index:
+                pos_new.append(new_batch.get_pos(None, isrc, index))
+                pos_old.append(self.get_pos(None, isrc, index))
+            getattr(new_batch, isrc)[pos_new] = getattr(self, isrc)[pos_old]
         return new_batch
 
     def trace_headers(self, header, flatten=False):
--- a/seismicpro/src/seismic_batch.py
+++ b/seismicpro/src/seismic_batch.py
@@ -644,10 +644,9 @@ class SeismicBatch(Batch):
         getattr(self, dst)[pos] = np.pad(data, **kwargs)
         return self
 
-    @action
     @inbatch_parallel(init="_init_component", target="threads")
     @apply_to_each_component
-    def sort_traces(self, index, *args, src, sort_by, dst=None):
+    def _sort(self, index, src, sort_by, sorting, dst=None):
         """Sort traces.
 
         Parameters
@@ -656,7 +655,40 @@ class SeismicBatch(Batch):
             The batch components to get the data from.
         dst : str, array-like
             The batch components to put the result in.
-        sort_by: str
+        sort_by : str
+            Sorting key.
+        sorting : str
+            Current sorting of `src` component
+
+        Returns
+        -------
+        batch : SeismicBatch
+            Batch with new trace sorting.
+        """
+        pos = self.get_pos(None, src, index)
+        df = self.index.get_df([index])
+
+        if sorting:
+            cols = [sorting, sort_by]
+            sorted_index_df = df[cols].sort_values(sorting)
+            order = np.argsort(sorted_index_df[sort_by].values)
+        else:
+            order = np.argsort(df[sort_by].values)
+
+        getattr(self, dst)[pos] = getattr(self, src)[pos][order]
+        return self
+
+    @action
+    def sort_traces(self, *args, src, sort_by, dst=None):
+        """Sort traces.
+
+        Parameters
+        ----------
+        src : str, array-like
+            The batch components to get the data from.
+        dst : str, array-like
+            The batch components to put the result in.
+        sort_by : str
             Sorting key.
 
         Returns
@@ -670,29 +702,17 @@ class SeismicBatch(Batch):
         else:
             sorting = None
 
-        pos = self.get_pos(None, src, index)
-        df = self.index.get_df([index])
+        if sorting == sort_by:
+            return self
 
-        if sorting:
-            if sorting == sort_by:
-                return self
-
-            cols = [sorting, sort_by]
-            sorted_index_df = df[cols].sort_values(sorting)
-            order = np.argsort(sorted_index_df[sort_by].values)
-        else:
-            order = np.argsort(df[sort_by].values)
-
-        getattr(self, dst)[pos] = getattr(self, src)[pos][order]
-
-        if pos == 0:
-            self.meta[dst]['sorting'] = sort_by
+        self._sort(src=src, sort_by=sort_by, sorting=sorting, dst=dst)
+        self.meta[dst]['sorting'] = sort_by
 
         return self
 
     @action
     @inbatch_parallel(init="indices", post='_post_filter_by_mask', target="threads")
-    def drop_zero_traces(self, index, src, num_zero, **kwargs):
+    def drop_zero_traces(self, index, src, num_zero, all_comps_sorted=True, **kwargs):
         """Drop traces with sequence of zeros longer than ```num_zero```.
 
         This action drops traces from index dataframe and from all batch components
@@ -704,6 +724,9 @@ class SeismicBatch(Batch):
             Size of the sequence of zeros.
         src : str, array-like
             The batch components to get the data from.
+        all_comps_sorted : bool
+            Check that all components have the same sorting to ensure that they are
+            modified in a same way.
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


seismicpro/src/seismic_batch.py
```diff
@@ -667,21 +665,36 @@ def sort_traces(self, index, *args, src, sort_by, dst=None):
             Batch with new trace sorting.
         """
         _ = args
+        sorting = self.meta[dst]['sorting']
```
nikita-klsh: is not it supposed to be `self.meta[src]` instead?

seismicpro/src/seismic_batch.py
```diff
@@ -667,21 +665,36 @@ def sort_traces(self, index, *args, src, sort_by, dst=None):
...
+
+        if sorting:
+            if sorting == sort_by:
+                return self
+
+            cols = [sorting, sort_by]
+            sorted_index_df = df[cols].sort_values(sorting)
+            order = np.argsort(sorted_index_df[sort_by].values)
+        else:
+            order = np.argsort(df[sort_by].tolist())
```
nikita-klsh: be more consistent and use either `.tolist()` or `.values`

seismicpro/src/seismic_batch.py
```diff
@@ -693,15 +706,45 @@ def drop_zero_traces(self, index, src, num_zero, **kwargs):
...
+            nonzero_indices = np.concatenate(([-1], nonzero_indices, [len(trace)]))
+            zero_seqs = np.diff(nonzero_indices) - 1
+            mask.append(np.max(zero_seqs) < num_zero)
+        mask = np.array(mask)
+
+        for comp in self.components:
+            getattr(self, comp)[pos] = getattr(self, comp)[pos][mask]
+
+        cols = [(INDEX_UID, src), (sorting, '')]
+        sorted_index_df = self.index.get_df(index)[cols].sort_values(sorting)
```
nikita-klsh:
```suggestion
        sorted_index_df = self.index.get_df([index])
```
In case `TraceIndex` you want the resulted slice to be `DataFrame`, not `Series`

nikita-klsh: yet not compatible with `TraceIndex`  as in this case `cols` contains not existing columns

seismicpro/src/seismic_batch.py
```diff
@@ -693,15 +706,45 @@ def drop_zero_traces(self, index, src, num_zero, **kwargs):
...
+        This action creates new instance of SeismicBatch with new index
+        instance.
         """
         _ = kwargs
+        sorting = self.meta[src]['sorting']
+        if sorting is None and isinstance(self.index, FieldIndex):
+            raise ValueError('traces in `{}` component should be sorted '
+                             'before dropping zero traces'.format(src))
+
+        has_same_sorting = all([self.meta[comp]['sorting'] == sorting for comp in self.components])
```
nikita-klsh: this might be a bottleneck as only the components initialized in the `index` constructor or generated by the actions decorated with `_init_component` are included in `self.meta`

seismicpro/src/seismic_batch.py
dpodvyaznikov: Yes, it is

seismicpro/src/seismic_batch.py
dpodvyaznikov: This is the exact behavior I wanted

seismicpro/src/seismic_batch.py
dpodvyaznikov: Indeed, for `TraceIndex` the logic should be different

dpodvyaznikov: It is now

seismicpro/src/seismic_batch.py
```diff
@@ -693,14 +709,46 @@ def drop_zero_traces(self, index, src, num_zero, **kwargs):
...
+        ValueError : if any component in batch has sorting different from `src`
+
+        Note
+        ----
+        This action creates new instance of SeismicBatch with new index
+        instance.
         """
         _ = kwargs
+        sorting = self.meta[src]['sorting']
+        if sorting is None and isinstance(self.index, FieldIndex):
```
nikita-klsh:
```suggestion
        if sorting is None and not isinstance(self.index, TraceIndex):
```
I think in case we use, for example, `Common Midpoint Index` or `Common Receiver index` we must still care about sorting

seismicpro/src/seismic_batch.py
```diff
@@ -175,30 +176,27 @@ def _post_filter_by_mask(self, mask, *args, **kwargs):
         Parameters
```
a-arefina: A bit misleading description, "union of all masks" can be interpreted as a set of masks from different conditions  for same list of  items

seismicpro/src/seismic_batch.py
```diff
@@ -211,11 +209,11 @@ def _post_filter_by_mask(self, mask, *args, **kwargs):
         for comp in batch.components:
             setattr(batch, comp, np.array([None] * len(batch.index)))
 
-        for i, index in enumerate(new_index):
+        for index in new_index:
             for isrc in batch.components:
-                pos = self.get_pos(None, isrc, index)
-                new_data = getattr(self, isrc)[pos][mask[pos]]
-                getattr(batch, isrc)[i] = new_data
+                pos_batch = batch.get_pos(None, isrc, index)
```
a-arefina: i would change `pos_batch` -> `pos_new` and  `batch`-> `new_batch`, `pos_self` -> `pos_old` or smth like this

seismicpro/src/seismic_batch.py
```diff
@@ -658,49 +655,125 @@ def sort_traces(self, index, *args, src, sort_by, dst=None):
             The batch components to get the data from.
         dst : str, array-like
             The batch components to put the result in.
-        sort_by: str
+        sort_by : str
             Sorting key.
+        sorting : str
```
a-arefina: rename to `current_sorting` or `current_sort_by`

seismicpro/src/seismic_batch.py
```diff
@@ -658,49 +655,125 @@ def sort_traces(self, index, *args, src, sort_by, dst=None):
...
-        order = np.argsort(df[sort_by].tolist())
+
+        if sorting:
+            cols = [sorting, sort_by]
+            sorted_index_df = df[cols].sort_values(sorting)
+            order = np.argsort(sorted_index_df[sort_by].values)
+        else:
+            order = np.argsort(df[sort_by].values)
+
         getattr(self, dst)[pos] = getattr(self, src)[pos][order]
```
a-arefina: what if we get `dst==None` here? If this is impossible or inevitably leads to error why make it default value?

seismicpro/src/seismic_batch.py
```diff
@@ -658,49 +655,125 @@ def sort_traces(self, index, *args, src, sort_by, dst=None):
...
-        if pos == 0:
-            self.meta[dst]['sorting'] = sort_by
+        return self
+
+    @action
+    def sort_traces(self, *args, src, sort_by, dst=None):
```
a-arefina: I would remove default value for `dst`

seismicpro/src/seismic_batch.py
```diff
@@ -658,49 +655,125 @@ def sort_traces(self, index, *args, src, sort_by, dst=None):
...
+        This action drops traces from index dataframe and from all batch components
+        according to the mask calculated on `src` component.
+
         Parameters
         ----------
         num_zero : int
             Size of the sequence of zeros.
         src : str, array-like
             The batch components to get the data from.
+        all_comps_sorted : bool
```
a-arefina: rename to `check_all_components_sorted`.
By the way, when do we need it False?

seismicpro/src/seismic_batch.py
```diff
@@ -658,49 +655,125 @@ def sort_traces(self, index, *args, src, sort_by, dst=None):
...
         """Drop traces with sequence of zeros longer than ```num_zero```.
 
+        This action drops traces from index dataframe and from all batch components
+        according to the mask calculated on `src` component.
+
         Parameters
         ----------
         num_zero : int
             Size of the sequence of zeros.
         src : str, array-like
```
a-arefina: I would rename this parameter, because here it has different meaning than `src` in all other actions

seismicpro/src/seismic_batch.py
```diff
@@ -658,49 +655,125 @@ def sort_traces(self, index, *args, src, sort_by, dst=None):
...
+        instance.
         """
         _ = kwargs
+        sorting = self.meta[src]['sorting']
+        if sorting is None and not isinstance(self.index, TraceIndex):
+            raise ValueError('traces in `{}` component should be sorted '
+                             'before dropping zero traces'.format(src))
+
+        if all_comps_sorted:
+            has_same_sorting = all([self.meta[comp]['sorting'] == sorting for comp in self.components])
```
a-arefina: you can remove list comprehension here and use generator expression

seismicpro/src/seismic_batch.py
```diff
@@ -658,49 +655,125 @@ def sort_traces(self, index, *args, src, sort_by, dst=None):
...
+            if not has_same_sorting:
+                raise ValueError('all components in batch should have same sorting')
```
a-arefina: do we need an error or resorting here?

seismicpro/src/seismic_batch.py
```diff
@@ -658,49 +655,125 @@ def sort_traces(self, index, *args, src, sort_by, dst=None):
...
+
         pos = self.get_pos(None, src, index)
         traces = getattr(self, src)[pos]
         mask = list()
-        for _, trace in enumerate(traces != 0):
-            diff_zeros = np.diff(np.append(np.where(trace)[0], len(trace)))
-            mask.append(False if len(diff_zeros) == 0 else np.max(diff_zeros) < num_zero)
+        for trace in traces:
+            nonzero_indices = np.nonzero(trace)[0]
```
a-arefina: `np.flatnonzero` or `np.nonzero(np.squeeze(trace))`, if we work trace-wise

seismicpro/src/seismic_batch.py
```diff
@@ -658,49 +655,125 @@ def sort_traces(self, index, *args, src, sort_by, dst=None):
...
-    def drop_zero_traces(self, index, src, num_zero, **kwargs):
+    def drop_zero_traces(self, index, src, num_zero, all_comps_sorted=True, **kwargs):
         """Drop traces with sequence of zeros longer than ```num_zero```.
 
+        This action drops traces from index dataframe and from all batch components
+        according to the mask calculated on `src` component.
+
         Parameters
         ----------
         num_zero : int
```
a-arefina: rename to `min_num_zeros` or smth like that and clarify the description

seismicpro/src/seismic_batch.py
```diff
@@ -658,49 +655,125 @@ def sort_traces(self, index, *args, src, sort_by, dst=None):
...
+            if not has_same_sorting:
+                raise ValueError('all components in batch should have same sorting')
+
         pos = self.get_pos(None, src, index)
         traces = getattr(self, src)[pos]
         mask = list()
-        for _, trace in enumerate(traces != 0):
-            diff_zeros = np.diff(np.append(np.where(trace)[0], len(trace)))
-            mask.append(False if len(diff_zeros) == 0 else np.max(diff_zeros) < num_zero)
+        for trace in traces:
```
a-arefina: Did you consider working with entire `traces` array here?

seismicpro/src/seismic_batch.py
```diff
@@ -658,49 +655,125 @@ def sort_traces(self, index, *args, src, sort_by, dst=None):
...
+            nonzero_indices = np.nonzero(trace)[0]
+            # add -1 and len(trace) indices to count leading and trailing zero sequences
+            nonzero_indices = np.concatenate(([-1], nonzero_indices, [len(trace)]))
+            zero_seqs = np.diff(nonzero_indices) - 1
+            mask.append(np.max(zero_seqs) < num_zero)
+        mask = np.array(mask)
+
+        for comp in self.components:
```
a-arefina: do you need this, if we apply _post_filter_by_mask?

seismicpro/src/seismic_batch.py
dpodvyaznikov: For now, there is no way for user to "drop" unnecessary components if there are any in batch. One may not care about sorting in those components, so this flag allows one to take responsibility for everything that happens in those unsorted components. 
After a few other modifications this behavior will be changed: user will provide a list of components to be preserved and they will be sorted in this action.

seismicpro/src/seismic_batch.py
dpodvyaznikov: It is possible.
This action is wrapped with `@apply_to_each_component`, which sets `dst` = `src` if `dst` is `None`.

seismicpro/src/seismic_batch.py
dpodvyaznikov: Argeed

seismicpro/src/seismic_batch.py
dpodvyaznikov: Will do

seismicpro/src/seismic_batch.py
dpodvyaznikov: Ok

seismicpro/src/seismic_batch.py
dpodvyaznikov: Done

seismicpro/src/seismic_batch.py
dpodvyaznikov: At this moment - error, later we'll add resorting.

seismicpro/src/seismic_batch.py
dpodvyaznikov: Yes, in this function we filter items in components, while in post function index is filtered

seismicpro/src/seismic_batch.py
dpodvyaznikov: What's the benefit of this?

seismicpro/src/seismic_batch.py
dpodvyaznikov: I need to get indices of nonzero elements, and they are of different length for each trace, so it is not possible to handle entire array

seismicpro/src/seismic_batch.py
dpodvyaznikov: Good idea

seismicpro/src/seismic_batch.py
dpodvyaznikov: Ok, I got it

seismicpro/src/seismic_batch.py
dpodvyaznikov: Changed description, not sure about renaming

seismicpro/src/seismic_batch.py
```diff
@@ -170,53 +171,52 @@ def _init_component(self, *args, dst, **kwargs):
...
-        batch.add_components(self.components)
-        batch.meta = self.meta
+        new_batch = type(self)(batch_index)
+        new_batch.add_components(self.components)
+        new_batch.meta = self.meta
 
-        for comp in batch.components:
-            setattr(batch, comp, np.array([None] * len(batch.index)))
+        for comp in new_batch.components:
+            setattr(new_batch, comp, np.array([None] * len(new_batch.index)))
```
roman-kh: new_batch.add_components(self.components, self.components * [new_batch,array_of_nones])

seismicpro/src/seismic_batch.py
```diff
@@ -170,53 +171,52 @@ def _init_component(self, *args, dst, **kwargs):
...
 
-        for i, index in enumerate(new_index):
-            for isrc in batch.components:
-                pos = self.get_pos(None, isrc, index)
-                new_data = getattr(self, isrc)[pos][mask[pos]]
-                getattr(batch, isrc)[i] = new_data
-        return batch
+        for index in new_index:
+            for isrc in new_batch.components:
```
roman-kh: It could be much more efficient to swap the cycles: for each component you can get all the data from the old batch and put it into the new batch all at once.

seismicpro/src/seismic_batch.py
dpodvyaznikov: Done

seismicpro/src/seismic_batch.py
dpodvyaznikov: Done

PR #14, opened 2019-12-04
Update setup

dpodvyaznikov:


seismicpro/src/seismic_batch.py
```diff
@@ -1377,3 +1377,48 @@ def shift_pick_phase(self, index, src, src_traces, dst=None, shift=1.5, threshol
...
         return self
+
+    @action
+    @inbatch_parallel(init='_init_component')
+    def _load_from_index(self, index, *args, key, dst, **kwargs):
+        """Load values from index dataframe to a component
+        """
+        _ = args, kwargs
+        pos = self.get_pos(None, None, index)
+        getattr(self, dst)[pos] = self.index.get_df(index)[key].values
```
nikita-klsh:
```suggestion
        getattr(self, dst)[pos] = self.index.get_df([index])[key].values
```
make it compatible with `TraceIndex`

seismicpro/src/seismic_batch.py
```diff
@@ -1377,3 +1377,48 @@ def shift_pick_phase(self, index, src, src_traces, dst=None, shift=1.5, threshol
         x += n_skip
         getattr(self, dst)[pos] = x
         return self
+
+    @action
+    @inbatch_parallel(init='_init_component')
+    def _load_from_index(self, index, *args, key, dst, **kwargs):
```
nikita-klsh: it would be nice if we could load multiple columns during single action call

seismicpro/src/seismic_batch.py
```diff
@@ -468,23 +471,18 @@ def _dump_picking(self, src, path, src_traces, input_units='samples', columns=('
...
-            columns = columns + (PICKS_FILE_HEADER, )
-
-        df = self.index.get_df(reset=False)
-        sort_by = self.meta.get(src, {}).get('sorting')
-        if sort_by is not None:
-            df = df.sort_values(by=sort_by)
-
-        df = df.loc[self.indices]
-        df = df.reset_index(drop=self.index.name is None)[list(columns)]
+        df = self.index.get_df(self.indices, reset=True)[list(columns)]
```
nikita-klsh:
```suggestion
        df = self.index.get_df()[list(columns)]
```
are not the arguments passed redundant?

seismicpro/src/seismic_batch.py
nikita-klsh: the action execution takes almost 1 sec for BS=64 and 10 sec for BS=1000.
I would suggest to rewrite it in non parallel manner, see `_load_picking`  method for details.

Issue #19, opened 2020-02-13
Can not iterate over seismograms if dataset contains multiple SEGYs


nikita-klsh:
`FieldRecord` is not unique ID anymore in this case

dpodvyaznikov:
This issue is addressed in #20

PR #20, opened 2020-03-05
Index algebra

nikita-klsh:
- Index algebra
- Update batchflow


seismicpro/src/seismic_index.py
```diff
@@ -183,6 +183,34 @@ def merge(self, x, **kwargs):
         indices = df.index.unique().sort_values()
         return type(self).from_index(index=indices, idf=df, index_name=self.name)
 
+    def concat(self, other):
+        """ Concatenate vertically current and `other` indices DataFrames.
```
a-arefina:
```suggestion
        """ Concatenate vertically current and `other` indices DataFrames.
```

dpodvyaznikov:
### Batch
* `equalize` now can invert equalization with `upscale` parameter.
--------------

### Index
* 'CDP', 'GroupX' and 'GroupY' are now default headers to load when index is created. Based on 'GroupX' and 'GroupY' a column named 'Group' is created, which can be used as receiver index.
* When `Index` instance is created from multiple files, their dataframes are concatenated. Now if dataframes have intersection in any of 'FieldRecord', 'Group' or 'CDP' columns, an error is raised.
* One still can concatenate multiple instances of `Index` using `concat` method or `+` operation. A warning about possible non-unique index values is shown in this case.
--------------

seismicpro/src/seismic_index.py
```diff
@@ -183,6 +184,37 @@ def merge(self, x, **kwargs):
...
+        """
+        warn_msg = 'Concatenation of several `Index` instances may result in non-unique index.'
+        warnings.warn(warn_msg, RuntimeWarning)
+
+        if not isinstance(other, (list, tuple)):
+            other = [other]
+
+        df = [self.get_df()]
+        other_dfs = [index.get_df() for index in other]
+        concat_dfs = pd.concat(df + other_dfs, ignore_index=True)
```
a-arefina, line 208: imo, this is more clear, because `df` is usually a dataframe, not a list of dataframes
```suggestion
        concat_dfs = pd.concat([self.get_df()] + other_dfs, ignore_index=True)
```

seismicpro/src/seismic_index.py
```diff
@@ -183,6 +184,37 @@ def merge(self, x, **kwargs):
...
+        ----------
+        other : Index or sequence of Index
+            Indices which dataframes are being concatenated.
+
+        Returns
+        -------
+            : TraceIndex
+            Concatenated index.
+        """
+        warn_msg = 'Concatenation of several `Index` instances may result in non-unique index.'
```
a-arefina, line 200: the message is not strictly correct: index of the resulting `Index` will be unique, because it is constructed to be such, but some entries will be missing if duplicated indices occur

seismicpro/src/utils.py
```diff
@@ -589,17 +591,25 @@ def make_segy_index(filename, extra_headers=None, limits=None):
...
+            del_headers = []
         elif extra_headers is None:
-            headers = DEFAULT_SEGY_HEADERS
+            headers = DEFAULT_SEGY_HEADERS + SUPPORT_SEGY_HEADERS
+            del_headers = SUPPORT_SEGY_HEADERS
         else:
-            headers = set(DEFAULT_SEGY_HEADERS + list(extra_headers))
+            extra_headers = extra_headers if isinstance(extra_headers, (list, tuple)) else [extra_headers]
+            headers = set(DEFAULT_SEGY_HEADERS + extra_headers + SUPPORT_SEGY_HEADERS)
+            del_headers = set(SUPPORT_SEGY_HEADERS).difference(set(extra_headers))
```
a-arefina, line 601: why not `set(SUPPORT_SEGY_HEADERS) - set(extra_headers)`?

seismicpro/src/utils.py
```diff
@@ -589,17 +591,25 @@ def make_segy_index(filename, extra_headers=None, limits=None):
...
 
         meta = dict()
 
         for k in headers:
             meta[k] = segyfile.attributes(getattr(segyio.TraceField, k))[limits]
 
         meta['file_id'] = np.repeat(filename, segyfile.tracecount)[limits]
+        meta['RecieverID'] = (0.5 * (meta['GroupX'] + meta['GroupY']) * \
+                              (meta['GroupX'] + meta['GroupY'] + 1) + meta['GroupY'])
+        for k in del_headers:
```
a-arefina, line 611: what about renaming `del_headers` to `del_meta`?

seismicpro/src/utils.py
```diff
@@ -589,17 +591,25 @@ def make_segy_index(filename, extra_headers=None, limits=None):
...
+            del_headers = set(SUPPORT_SEGY_HEADERS).difference(set(extra_headers))
 
         meta = dict()
 
         for k in headers:
             meta[k] = segyfile.attributes(getattr(segyio.TraceField, k))[limits]
 
         meta['file_id'] = np.repeat(filename, segyfile.tracecount)[limits]
+        meta['RecieverID'] = (0.5 * (meta['GroupX'] + meta['GroupY']) * \
+                              (meta['GroupX'] + meta['GroupY'] + 1) + meta['GroupY'])
```
a-arefina, line 610: Why this form? It doesn't produce unique values for different X and Y

seismicpro/src/utils.py
```diff
@@ -628,6 +638,11 @@ def build_segy_df(extra_headers=None, name=None, limits=None, **kwargs):
     index = FilesIndex(**kwargs)
     df = pd.concat([make_segy_index(index.get_fullpath(i), extra_headers, limits) for
                     i in sorted(index.indices)])
+    if len(index) > 1:
+        for colname in GATHER_HEADERS:
+            if any(df[[colname, 'file_id']].groupby(colname).nunique()[('file_id')] > 1):
```
a-arefina, line 643:
```suggestion
            if any(df[[colname, 'file_id']].groupby([colname, 'file_id']).size()> 1):
```

seismicpro/src/seismic_batch.py
```diff
@@ -1519,23 +1522,23 @@ def equalize(self, index, src, dst, params, survey_id_col=None):
         pair.
         3. This action copies all meta from `src` component to `dst` component.
         """
-        if not isinstance(self.index, FieldIndex):
-            raise ValueError("Index must be FieldIndex, not {}".format(type(self.index)))
+        if not isinstance(self.index, (FieldIndex, TraceIndex)):
```
a-arefina, line 1525: `CustomIndex`, ex. for `CDP` or `RecieverID`, is also OK

seismicpro/src/utils.py
```diff
@@ -589,17 +591,25 @@ def make_segy_index(filename, extra_headers=None, limits=None):
...
         if extra_headers == 'all':
             headers = [h.__str__() for h in segyio.TraceField.enums()]
+            del_headers = []
         elif extra_headers is None:
-            headers = DEFAULT_SEGY_HEADERS
+            headers = DEFAULT_SEGY_HEADERS + SUPPORT_SEGY_HEADERS
+            del_headers = SUPPORT_SEGY_HEADERS
         else:
-            headers = set(DEFAULT_SEGY_HEADERS + list(extra_headers))
+            extra_headers = extra_headers if isinstance(extra_headers, (list, tuple)) else [extra_headers]
```
a-arefina, line 599: array-likes are converted to lists, so that `+` op works in the next line
```suggestion
            extra_headers = [extra_headers] if isinstance(extra_headers, str) else list(extra_headers)
```

seismicpro/src/utils.py
dpodvyaznikov: All traces in a shot gather share same `FieldRecord`-`file_id` pair, so `size` of such a group would be definitely more than one.

seismicpro/src/utils.py
dpodvyaznikov: Changed to `tmp_headers`

seismicpro/src/seismic_index.py
dpodvyaznikov: It may not be strict, but it is very neat and clear imo

a-arefina: neat, and clean, and wrong :)

what about 'If several `Index` instances contain duplicated index values, only one of them will be left after concatenation.'?

seismicpro/src/utils.py
```diff
@@ -589,17 +591,24 @@ def make_segy_index(filename, extra_headers=None, limits=None):
...
+            tmp_headers = []
         elif extra_headers is None:
-            headers = DEFAULT_SEGY_HEADERS
+            headers = DEFAULT_SEGY_HEADERS + SUPPORT_SEGY_HEADERS
+            tmp_headers = SUPPORT_SEGY_HEADERS
         else:
-            headers = set(DEFAULT_SEGY_HEADERS + list(extra_headers))
+            extra_headers = [extra_headers] if isinstance(extra_headers, str) else list(extra_headers)
+            headers = set(DEFAULT_SEGY_HEADERS + extra_headers + SUPPORT_SEGY_HEADERS)
+            tmp_headers = set(SUPPORT_SEGY_HEADERS) - (set(extra_headers))
```
a-arefina, line 601:
```suggestion
            tmp_headers = set(SUPPORT_SEGY_HEADERS) - set(extra_headers)
```

a-arefina (approved):
Minor fixes in suggestions + re-write warning

dpodvyaznikov:
Closes #19

PR #22, opened 2020-03-27
Rec index

dpodvyaznikov:
Add `RecIndex` to iterate over receiver gathers.

### DIFF
--- a/seismicpro/src/seismic_index.py
+++ b/seismicpro/src/seismic_index.py
@@ -438,6 +438,9 @@ class BinsIndex(TraceIndex):
 
 class RecIndex(TraceIndex):
     """ Index for recievers.
+    It uses `Group` columns of the dataframe as index. `Group` is built by concatenation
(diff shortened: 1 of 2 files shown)


seismicpro/src/seismic_index.py
```diff
@@ -434,3 +434,34 @@ def show_heatmap(self, **kwargs):
             show_2d_heatmap(self._idf, **kwargs)
         else:
             show_1d_heatmap(self._idf, **kwargs)
+
+
+class RecIndex(TraceIndex):
```
a-arefina: Let's finally extract SeismicIndex class

nikita-klsh:
Add `RecIndex` to `__init__.py`

seismicpro/src/seismic_index.py
a-arefina: OK, I will do it myself in my next PR into SeismicPro

dpodvyaznikov:
Unnecessary right now due to planned changes in index.

PR #24, opened 2020-04-13
add copy_meta action

anton-br:
This pull request is aimed to add an oppotunity to copy meta information from one component to another. Only one action `copy_meta` contains here.

We need this because there are actions that use meta data inside:
* _dump_split_segy
* _dump_picking
* hodograph_straightening
* calculate_semblance

Without coping meta we aren't able to use them with any components, we can use them only will component with loaded data inside (usually 'raw').

### DIFF
--- a/seismicpro/src/seismic_batch.py
+++ b/seismicpro/src/seismic_batch.py
@@ -284,7 +284,7 @@ class SeismicBatch(Batch):
             'new_raw': {'arg_1': 'val_1', 'arg_2': 'val_2'}}
 
 
-        Use for few components:
+        Use for several components:
             >>> batch.copy_meta(['raw', 'raw'], ['new_raw', 'new_raw_2'], key=[['arg_1'], ['arg_2']])
             >>> batch.meta
             {'raw': {'arg_1': 'val_1', 'arg_2': 'val_2', 'arg_3': 'val_3'},
--- a/seismicpro/src/seismic_batch.py
+++ b/seismicpro/src/seismic_batch.py
@@ -241,6 +241,82 @@ class SeismicBatch(Batch):
         return np.array(np.split(values, np.cumsum(tracecounts)[:-1]) + [None])[:-1]
 
     @action
+    def copy_meta(self, from_comp, to_comp, keys=None):
+        """Copy meta from one component to another or form few components. One can copy either
+        full meta or only particular keys.
+
+        Parameters
+        ----------
+        from_comp : str or array-like
+            Component's name to copy meta from or list of component's names.
+        to_comp : str or array-like
+            Component's name to copy meta in or list of component's names.
+        keys : str, array-like or None, optional
+            if None, all meta will be copied
+            if str, only meta[keys] will be copied
+            if array-like,
+                if 1d array-like, for each `to_comp` will be used same keys
+                if 2d array-like, for each `to_comp` will be used particular keys.
+                It means that i-th component will contain only `keys[i]` elements from meta.
+
+        Raises
+        ------
+        ValueError : if `from_comp` and `to_comp` have different length.
+        ValueError : if one of given to `from_comp` component doesn't exist.
+
+        Note
+        ----
+            The copied meta will not overwrite the old meta.
+
+        Examples
+        --------
+        Use all meta:
+            >>> batch.copy_meta('raw', 'new_raw')
+            >>> batch.meta
+            {'raw': {'arg_1': 'val_1', 'arg_2': 'val_2', 'arg_3': 'val_3'},
+            'new_raw': {'arg_1': 'val_1', 'arg_2': 'val_2', 'arg_3': 'val_3'}}
+
+
+        Copy two values:
+            >>> batch.copy_meta('raw', 'new_raw', key=['arg_1', 'arg_2'])
+            >>> batch.meta
+            {'raw': {'arg_1': 'val_1', 'arg_2': 'val_2', 'arg_3': 'val_3'},
+            'new_raw': {'arg_1': 'val_1', 'arg_2': 'val_2'}}
+
+
+        Use for few components:
+            >>> batch.copy_meta(['raw', 'raw'], ['new_raw', 'new_raw_2'], key=[['arg_1'], ['arg_2']])
+            >>> batch.meta
+            {'raw': {'arg_1': 'val_1', 'arg_2': 'val_2', 'arg_3': 'val_3'},
+            'new_raw': {'arg_1': 'val_1'},
+            'new_raw': {'arg_2': 'val_2'}}
+        """
+        from_comp = (from_comp, ) if isinstance(from_comp, str) else from_comp
+        to_comp = (to_comp, ) if isinstance(to_comp, str) else to_comp
+
+        if len(from_comp) != len(to_comp):
+            raise ValueError('Length from_comp should be equal to to_comp length.')
+
+        if isinstance(keys, str) or keys is None:
+            keys = (keys, ) * len(from_comp)
+
+        for ix, (fr_comp, t_comp) in enumerate(zip(from_comp, to_comp)):
+            if fr_comp not in self.meta:
+                raise ValueError(f'{fr_comp} not exist.')
+
+            added_meta = self.meta[fr_comp]
+            if keys[ix] is not None:
+                if isinstance(keys[ix], str):
+                    added_meta = {k: added_meta[k] for k in keys}
+                else:
+                    added_meta = {k: added_meta[k] for k in keys[ix]}
+            new_meta = dict(added_meta)
+            new_meta.update(**self.meta[t_comp])
+            self.meta[t_comp] = new_meta
+
+        return self
+
+    @action
     @inbatch_parallel(init="_init_component", target="threads")
     @apply_to_each_component
     def apply_along_axis(self, index, func, *args, src, dst=None, slice_axis=0, **kwargs):
--- a/seismicpro/src/seismic_batch.py
+++ b/seismicpro/src/seismic_batch.py
@@ -304,15 +304,14 @@ class SeismicBatch(Batch):
             if fr_comp not in self.meta:
                 raise ValueError(f'{fr_comp} not exist.')
 
-            added_meta = self.meta[fr_comp]
+            added_meta = self.meta[fr_comp].copy()
             if keys[ix] is not None:
                 if isinstance(keys[ix], str):
                     added_meta = {k: added_meta[k] for k in keys}
                 else:
                     added_meta = {k: added_meta[k] for k in keys[ix]}
-            new_meta = dict(added_meta)
-            new_meta.update(**self.meta[t_comp])
-            self.meta[t_comp] = new_meta
+            added_meta.update(**self.meta[t_comp])
+            self.meta[t_comp] = added_meta
 
         return self
 
--- a/seismicpro/src/seismic_batch.py
+++ b/seismicpro/src/seismic_batch.py
@@ -283,8 +283,8 @@ class SeismicBatch(Batch):
 
             if self.meta[t_comp]:
                 warnings.warn("Meta of component {} is not empty and".format(t_comp) + \
-                                " will be replaced by the meta from component {}.".format(fr_comp),
-                                UserWarning)
+                              " will be replaced by the meta from component {}.".format(fr_comp),
+                              UserWarning)
             self.meta[t_comp] = self.meta[fr_comp].copy()
 
         return self
--- a/seismicpro/src/seismic_batch.py
+++ b/seismicpro/src/seismic_batch.py
@@ -1,5 +1,6 @@
 """Seismic batch.""" # pylint: disable=too-many-lines
 import os
+import warnings
 from textwrap import dedent
 import numpy as np
 import matplotlib.pyplot as plt
@@ -160,28 +161,32 @@ class SeismicBatch(Batch):
         if preloaded is None:
             self.meta = dict()
 
-    def _init_component(self, *args, dst, **kwargs):
+    def _init_component(self, src, dst, *args, **kwargs):
         """Create and preallocate a new attribute with the name ``dst`` if it
         does not exist and return batch indices."""
         _ = args
-        src = kwargs.get('src')
         dst = (dst, ) if isinstance(dst, str) else dst
+        copy_meta = True
 
         if isinstance(src, str):
             src = (src, ) * len(dst)
+
         elif len(dst) != len(src):
-            raise ValueError(f'Length of src and dst should be equal. Now is len(src)={len(src)}, '\
-                             f'len(dst)={len(dst)}')
+            copy_meta = False
+            warnings.warn("length of src and dst are not equal, note that meta for new components will be empty.")
 
-        for comp_from, comp_to in zip(src, dst):
-            if comp_to not in self.meta:
-                if comp_from in self.meta:
-                    self.meta[comp_to] = self.meta[comp_from].copy()
-                else:
-                    self.meta[comp_to] = dict()
-
-            if self.components is None or comp_to not in self.components:
-                self.add_components(comp_to, init=self.array_of_nones)
+        for ix, idst in enumerate(dst):
+            if copy_meta:
+                isrc = src[ix]
+                if idst not in self.meta:
+                    if isrc in self.meta:
+                        self.meta[idst] = self.meta[isrc].copy()
+                    else:
+                        self.meta[idst] = dict()
+            else:
+                self.meta[idst] = dict()
+            if self.components is None or idst not in self.components:
+                self.add_components(idst, init=self.array_of_nones)
 
         return self.indices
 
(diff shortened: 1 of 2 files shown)


seismicpro/src/seismic_batch.py
```diff
@@ -240,6 +240,82 @@ def trace_headers(self, header, flatten=False):
...
+
+
+        Copy two values:
+            >>> batch.copy_meta('raw', 'new_raw', key=['arg_1', 'arg_2'])
+            >>> batch.meta
+            {'raw': {'arg_1': 'val_1', 'arg_2': 'val_2', 'arg_3': 'val_3'},
+            'new_raw': {'arg_1': 'val_1', 'arg_2': 'val_2'}}
+
+
+        Use for few components:
```
a-arefina, line 287:
```suggestion
        Use for several components:
```

seismicpro/src/seismic_batch.py
```diff
@@ -240,6 +240,82 @@ def trace_headers(self, header, flatten=False):
...
+            >>> batch.copy_meta(['raw', 'raw'], ['new_raw', 'new_raw_2'], key=[['arg_1'], ['arg_2']])
```
a-arefina, line 288: Do we really need this mode? In this case to separate copy_meta actions would be more readable, imo

seismicpro/src/seismic_batch.py
```diff
@@ -240,6 +240,82 @@ def trace_headers(self, header, flatten=False):
...
+        ValueError : if one of given to `from_comp` component doesn't exist.
+
+        Note
+        ----
+            The copied meta will not overwrite the old meta.
+
+        Examples
+        --------
+        Use all meta:
+            >>> batch.copy_meta('raw', 'new_raw')
```
a-arefina, line 274: It would be great if you added an example of meta copying in one of the cases described above to the batch tutorial. I don't insist on this, though

seismicpro/src/seismic_batch.py
```diff
@@ -240,6 +240,82 @@ def trace_headers(self, header, flatten=False):
...
+        from_comp = (from_comp, ) if isinstance(from_comp, str) else from_comp
+        to_comp = (to_comp, ) if isinstance(to_comp, str) else to_comp
+
+        if len(from_comp) != len(to_comp):
+            raise ValueError('Length from_comp should be equal to to_comp length.')
+
+        if isinstance(keys, str) or keys is None:
+            keys = (keys, ) * len(from_comp)
+
+        for ix, (fr_comp, t_comp) in enumerate(zip(from_comp, to_comp)):
```
nikita-klsh, line 303: Why not use `@apply_to_each_component` decorator?

seismicpro/src/seismic_batch.py
anton-br: i think it is usefull if you want to copy one or two meta's keys from different components. In this case it won't lose clarity.

seismicpro/src/seismic_batch.py
```diff
@@ -240,6 +240,82 @@ def trace_headers(self, header, flatten=False):
...
+            if fr_comp not in self.meta:
+                raise ValueError(f'{fr_comp} not exist.')
+
+            added_meta = self.meta[fr_comp]
```
nikita-klsh, line 307:
```suggestion
            added_meta = self.meta[fr_comp].copy()
```
Looks like you can avoid using intermediate dict `new_meta` at all by updating directly `added_meta` bellow

anton-br:
> Are you sure that this must be explicitly used as action in the pipelines, not just an ordinary method in other actions?

How will you understang what do you want to copy and from which component?
I thought this way would be the clearest way to do it.

seismicpro/src/seismic_batch.py
anton-br: Because we dont need to always copy meta for each component in pipeline. Usually we need to copy meta only in one (or few) components.

nikita-klsh: It does exactly this `for` loop when there are few components

seismicpro/src/seismic_batch.py
```diff
@@ -163,14 +163,25 @@ def __init__(self, index, *args, preloaded=None, **kwargs):
...
         dst = (dst, ) if isinstance(dst, str) else dst
 
-        for comp in dst:
-            self.meta[comp] = self.meta[comp] if comp in self.meta else dict()
-
-            if self.components is None or comp not in self.components:
-                self.add_components(comp, init=self.array_of_nones)
+        if isinstance(src, str):
+            src = (src, ) * len(dst)
+        elif len(dst) != len(src):
```
nikita-klsh, line 172: `@apply_to_each_component` is more appropriate place for this kind of check

anton-br: right now `@apply_to_each_component` can't works with returned indices. So we have a two chairs here:
1. We will have a quite similar parts of code in `@apply_to_each_comoponent` and in `_init_component`.
2. I will rewite `@apply_to_each_component` a little bit and we might meet some errors in future because I can't check all the functions for reproducibility. (Also we will call decoratior while we call another decorator)

Which one do you like best?)

a-arefina: `@apply_to_each_components` handles the case when `len(src) == len(dst)`. This function is broader an can be used with any combination of `len(src), len(dst)`. 

We can say where to get initial meta for each `dst` only when `len(src) == len(dst)` or `len(src) == 1`, but other cases are possible. So in case `len(src) != len(dst)` we just shouldn't copy meta implicitly. 

We shouldn't throw any error when `len(src) != len(dst)`, because it can be a normal situation

seismicpro/src/seismic_batch.py
```diff
@@ -163,14 +163,25 @@ def __init__(self, index, *args, preloaded=None, **kwargs):
     def _init_component(self, *args, dst, **kwargs):
         """Create and preallocate a new attribute with the name ``dst`` if it
         does not exist and return batch indices."""
-        _ = args, kwargs
+        _ = args
+        src = kwargs.get('src')
```
a-arefina, line 167: can we add `src` to  `_init_component` arguments explicitly?

seismicpro/src/seismic_batch.py
anton-br: @nikita-klsh , if you are agree, i will close this conversation. Right now i've added case when `len(src) != len(dst)` so we can't use `@apply_to_each_components`  decorator here.

seismicpro/src/seismic_batch.py
```diff
@@ -160,17 +161,32 @@ def __init__(self, index, *args, preloaded=None, **kwargs):
...
+            if copy_meta:
+                isrc = src[ix]
+                if idst not in self.meta:
+                    if isrc in self.meta:
+                        self.meta[idst] = self.meta[isrc].copy()
+                    else:
+                        self.meta[idst] = dict()
+            else:
+                self.meta[idst] = dict()
+            if self.components is None or idst not in self.components:
```
a-arefina, line 188: add blank line above please

tutorials/2.Batch.ipynb
```diff
@@ -1477,7 +1477,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.5.2"
+   "version": "3.6.9"
```
a-arefina, line 1480: Did you recalculate tutorials to ensure everything still works?

anton-br: no, i didn't. This change make no sense, thus i will return old version.

seismicpro/src/seismic_batch.py
```diff
@@ -160,17 +161,33 @@ def __init__(self, index, *args, preloaded=None, **kwargs):
         if preloaded is None:
             self.meta = dict()
 
-    def _init_component(self, *args, dst, **kwargs):
+    def _init_component(self, src, dst, *args, **kwargs):
```
dpodvyaznikov, line 164:
```suggestion
    def _init_component(self, *args, src, dst, **kwargs):
```
This way `src` and `dst` forced to be keyword

seismicpro/src/seismic_batch.py
```diff
@@ -160,17 +161,33 @@ def __init__(self, index, *args, preloaded=None, **kwargs):
...
-        for comp in dst:
-            self.meta[comp] = self.meta[comp] if comp in self.meta else dict()
+        if isinstance(src, str):
+            src = (src, ) * len(dst)
+
+        elif len(dst) != len(src):
+            copy_meta = False
+            warnings.warn("length of src and dst are not equal, note that meta for new components will be empty.")
+
+        for ix, idst in enumerate(dst):
```
dpodvyaznikov, line 178:
```suggestion
        for isrc, idst in zip(src, dst):
```

anton-br: you can't use `zip(src, dst)` iff len(src) and len(dst) are equal but it is not always true.

seismicpro/src/seismic_batch.py
anton-br: > It seems that proposed changes "as is" will break `SeismicBatch.load`.
> 
> It calls `_load_segy(src=components, dst=components, ...)`, which calls `_load_from_segy_file`, wrapped with `inbatch_parallel(init="_init_component", ...)`. This way `_init_component` will get yet nonexistent component as `src`.

function `_load_from_segy_file` calls with `(src=components, dst=components, tslice=tslice)` as an arguments. So in this case it will works properly as well. As far as i understand we always have `src` component when use  `@inbatch_parallel` decorator.

seismicpro/src/seismic_batch.py
```diff
@@ -240,6 +240,39 @@ def trace_headers(self, header, flatten=False):
 
         return np.array(np.split(values, np.cumsum(tracecounts)[:-1]) + [None])[:-1]
 
+    def copy_meta(self, from_comp, to_comp):
+        """Copy meta from one component to another or form few components. One can copy either
```
dpodvyaznikov, line 244: Add empty lines between sections in docstring

seismicpro/src/seismic_batch.py
```diff
@@ -240,6 +240,39 @@ def trace_headers(self, header, flatten=False):
...
+            Component's name to copy meta from or list of component's names.
+        to_comp : str or array-like
+            Component's name to copy meta in or list of component's names.
+        Raises
+        ------
+            ValueError : if `from_comp` and `to_comp` have different length.
+            ValueError : if one of given to `from_comp` component doesn't exist.
+        Note
+        ----
+            The copied meta will not overwrite the old meta.
```
dpodvyaznikov, line 258: Why not make an `overwrite` parameter with default `False`?

dpodvyaznikov:
I understand correctly that the tutorial wasn't modified?

anton-br:
> I understand correctly that the tutorial wasn't modified?
it is not the action anymore. So yes, there is nothing to show in the tutorial.. (or is there?)

seismicpro/src/seismic_batch.py
anton-br: Do you have an example where you want to overwrite the meta?

seismicpro/src/seismic_batch.py
```diff
@@ -240,6 +240,44 @@ def trace_headers(self, header, flatten=False):
...
+        """
+        from_comp = (from_comp, ) if isinstance(from_comp, str) else from_comp
+        to_comp = (to_comp, ) if isinstance(to_comp, str) else to_comp
+
+        if len(from_comp) != len(to_comp):
+            raise ValueError('Length from_comp should be equal to to_comp length.')
+
+        for fr_comp, t_comp in zip(from_comp, to_comp):
+            if fr_comp not in self.meta:
+                raise ValueError(f'{fr_comp} not exist.')
```
a-arefina, line 271: f-strings are cool, but SeismicPro supports python 3.5, so we can't use them :(

seismicpro/src/seismic_batch.py
```diff
@@ -749,7 +802,7 @@ def sort_traces(self, *args, src, sort_by, dst=None):
 
         self._sort(src=src, sort_by=sort_by, current_sorting=current_sorting, dst=dst)
         self.meta[dst]['sorting'] = sort_by
-
+        self.copy_meta(src, dst)
```
a-arefina, line 805: do this before setting `sort_by` in new meta

seismicpro/src/seismic_batch.py
a-arefina, line 271:
```suggestion
                raise ValueError('{} does not exist.'.format(fr_comp))
```

seismicpro/src/seismic_batch.py
```diff
@@ -240,6 +240,44 @@ def trace_headers(self, header, flatten=False):
...
+
+        Note
+        ----
+        The copied meta will not overwrite the old meta.
+        """
+        from_comp = (from_comp, ) if isinstance(from_comp, str) else from_comp
+        to_comp = (to_comp, ) if isinstance(to_comp, str) else to_comp
+
+        if len(from_comp) != len(to_comp):
+            raise ValueError('Length from_comp should be equal to to_comp length.')
```
a-arefina, line 267: rewrite the error msg please

seismicpro/src/seismic_batch.py
```diff
@@ -240,6 +240,44 @@ def trace_headers(self, header, flatten=False):
 
         return np.array(np.split(values, np.cumsum(tracecounts)[:-1]) + [None])[:-1]
 
+    def copy_meta(self, from_comp, to_comp, overwrite=False):
+        """Copy meta from one component to another or form few components. One can copy either
+        full meta or only particular keys.
```
a-arefina, line 245: please remove the phrase about copying certain keys

seismicpro/src/seismic_batch.py
```diff
@@ -240,6 +240,44 @@ def trace_headers(self, header, flatten=False):
 
         return np.array(np.split(values, np.cumsum(tracecounts)[:-1]) + [None])[:-1]
 
+    def copy_meta(self, from_comp, to_comp, overwrite=False):
```
a-arefina, line 243: what about making `overwrite=True` the default?

anton-br: I think that in most of cases we don't want to copy the meta

dpodvyaznikov: I'd suggest to remove the default at all. This is something you should think through when using this method, so it is better to be explicit.

anton-br: ok 🤔

anton-br:
> And merge master into your branch.

I think is aleady up to data. Please tell me if i wrong but i can't merge it :c

seismicpro/src/seismic_batch.py
```diff
@@ -240,6 +240,45 @@ def trace_headers(self, header, flatten=False):
 
         return np.array(np.split(values, np.cumsum(tracecounts)[:-1]) + [None])[:-1]
 
+    def copy_meta(self, from_comp, to_comp, overwrite):
+        """Copy meta from one component to another or form list of components to list of
```
nikita-klsh, line 244: is it form or from?

seismicpro/src/seismic_batch.py
nikita-klsh: i can barely imagine cases when we want NOT to overwrite meta
there is a lot of hardcoded `overwrite=True` flags in actions, setting default value would allow to avoid this.

seismicpro/src/seismic_batch.py
anton-br: fixed

seismicpro/src/seismic_batch.py
anton-br: We don't want to verwrite the meta in `crop` action for example. For me also `overwrite=True` looks like better solution (change my mind after previous comment). We can change it if everyone supports it. (@a-arefina @dpodvyaznikov )

anton-br:
This pull request has never been so close to merging, but:
Due to the thread about `overwrite` parameter and after discussion with @dpodvyaznikov  we descided to reconstruct the approach of copying meta data following way. Now, we will always replace meta from `src` in`dst` (doesn't meter whether `dst` empty or not).  In our point of view, this approach is more straightforward and clear. Looking forward to your comments @a-arefina @nikita-klsh @dpodvyaznikov .

Let the games begin. (again)

seismicpro/src/seismic_batch.py
```diff
@@ -240,6 +241,54 @@ def trace_headers(self, header, flatten=False):
...
         return np.array(np.split(values, np.cumsum(tracecounts)[:-1]) + [None])[:-1]
 
+    def copy_meta(self, from_comp, to_comp):
+        """Copy meta from one component to another or from list of components to list of
+        components with same length.
+
+        Parameters
+        ----------
+        from_comp : str or array-like
+            Component's name to copy meta from or list of component's names.
```
dpodvyaznikov, line 251:
```suggestion
            Component's name to copy meta from or list of components' names.
```

seismicpro/src/seismic_batch.py
```diff
@@ -240,6 +241,54 @@ def trace_headers(self, header, flatten=False):
...
+        to_comp : str or array-like
+            Component's name to copy meta in or list of component's names.
```
dpodvyaznikov, line 253:
```suggestion
            Component's name to copy meta in or list of components' names.
```

PR #25, opened 2020-04-29
Assemble crops

a-arefina:
action to assemble crops


a-arefina:
[assemble_crop](https://trello.com/c/ZihYNJD8/2-assemblecrop)

seismicpro/src/seismic_batch.py
```diff
@@ -1485,7 +1570,93 @@ def crop(self, index, src, coords, shape, dst=None):
...
+            raise NotImplementedError("Index can't be SegyFilesIndex")
+
+        pos = self.get_pos(None, None, index)
+        field = getattr(self, src)[pos]
+
+        len_x, len_y = field.shape
+        x, y = shape
+
+        coords_x = np.arange(0, len_x, x)
+        if len_x % x != 0 and drop_last:
```
anton-br, line 1652: why don't you use zero_pad here as well?

seismicpro/src/seismic_batch.py
```diff
@@ -1075,6 +1077,74 @@ def seismic_plot(self, src, index, wiggle=False, xlim=None, ylim=None, std=1, #
...
+        pos = self.get_pos(None, 'indices', index)
+
+        if src_picking is not None:
+            raise NotImplementedError()
+
+        pts_picking = None
+
+        arrs = getattr(self, src)[pos]
+        total_crops = len(arrs)
+        names = self.meta[src]['coords'][index]
```
anton-br, line 1130: In my opinion, you should check whether this key exist in meta or not.

seismicpro/src/seismic_batch.py
```diff
@@ -1075,6 +1077,74 @@ def seismic_plot(self, src, index, wiggle=False, xlim=None, ylim=None, std=1, #
...
+        Multi-column subplots.
+        """
+        pos = self.get_pos(None, 'indices', index)
+
+        if src_picking is not None:
+            raise NotImplementedError()
+
+        pts_picking = None
+
+        arrs = getattr(self, src)[pos]
```
anton-br, line 1128: rename `arrs` please

seismicpro/src/seismic_batch.py
```diff
@@ -1075,6 +1077,74 @@ def seismic_plot(self, src, index, wiggle=False, xlim=None, ylim=None, std=1, #
...
+                   wiggle=False, std=1,
+                   src_picking=None, s=None, scatter_color=None,
+                   figsize=None, title=None,
+                   save_to=None, dpi=None, **kwargs):
+        """Plot seismic traces.
+
+        Parameters
+        ----------
+        src : str or array of str
+            The batch component(s) with data to show.
```
anton-br, line 1091:
```suggestion
            The batch component(s) with crops to show.
```

seismicpro/src/seismic_batch.py
```diff
@@ -1075,6 +1077,74 @@ def seismic_plot(self, src, index, wiggle=False, xlim=None, ylim=None, std=1, #
...
+        if src_picking is not None:
+            raise NotImplementedError()
+
+        pts_picking = None
+
+        arrs = getattr(self, src)[pos]
+        total_crops = len(arrs)
+        names = self.meta[src]['coords'][index]
+
+        if num_crops is not None and num_crops < len(arrs):
```
anton-br, line 1132: Correct me if I'm wrong. But can't we have a desire to look at specific cropes?  Or is only the number of drops an important parameter?

seismicpro/src/seismic_batch.py
```diff
@@ -1075,6 +1077,74 @@ def seismic_plot(self, src, index, wiggle=False, xlim=None, ylim=None, std=1, #
...
+                   save_to=None, dpi=None, **kwargs):
+        """Plot seismic traces.
+
+        Parameters
+        ----------
+        src : str or array of str
+            The batch component(s) with data to show.
+        index : same type as batch.indices
+            Data index to show.
+        num_crops: int or None
```
anton-br, line 1094: may be set some small number of crops as default (3-5 for example)? looks like if there are more than 20 crops, it will be useless to show them all...

seismicpro/src/seismic_batch.py
```diff
@@ -1485,7 +1570,93 @@ def crop(self, index, src, coords, shape, dst=None):
...
+        """
+        Assembles crops from `src` into single seismogram using coordinates from `coords`
+
+        Parameters
+        ----------
+        src : str
+            component with crops
+        dst : str
+            component to put the result to. If None, `dst` == `src`
+        coords : list of tuples (int, int)
```
anton-br, line 1592: there are no such argument for this function. Please add to the docs that this parameter you get from meta.

seismicpro/src/seismic_batch.py
```diff
@@ -1485,7 +1570,93 @@ def crop(self, index, src, coords, shape, dst=None):
...
+            list of coordinates of upper left corners of crops from `src`.
+            The length of the list should be same as number of crops in `src`
+        """
+
+        if isinstance(self.index, SegyFilesIndex):
```
anton-br, line 1597: I've never use `SegyFileIndex` while work with already loaded data. Do we really need this check here?

seismicpro/src/seismic_batch.py
a-arefina: `arrs` is the name of corresponding `seismic_plot` parameter, so I would leave this name

seismicpro/src/seismic_batch.py
a-arefina: Yes, but if we have a small number of crops we might want to see them all without having to specify the exact number of crops when it is more than the default value

seismicpro/src/seismic_batch.py
a-arefina: The only way to specify what crops to show that I can think of is to pass their upper-left corner coordinates. We'll have to check if such crops exist and select them. It seems rather complicated so I would leave this until a real need to look at specific crops emerges.

seismicpro/src/seismic_batch.py
a-arefina: Because `drop_last` is the option to drop incomplete crops

seismicpro/src/seismic_batch.py
```diff
@@ -1485,7 +1570,93 @@ def crop(self, index, src, coords, shape, dst=None):
...
+            crop_counts[x:x1, y:y1] += 1
+
+        crop_counts[crop_counts == 0] = 1
+
+        getattr(self, dst)[pos] = res / crop_counts
+
+    @action
+    @inbatch_parallel(init='_init_component')
+    @apply_to_each_component
+    def make_grid_for_crops(self, index, src, dst, shape, drop_last=True):
```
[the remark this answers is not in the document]
a-arefina: This function is useful only during inference, we'll need other cropping strategies while training, so I don't feel that this should be the default

seismicpro/src/seismic_batch.py
a-arefina: If we called crop action on the current batch, we have all those keys, otherwise plotting crops is useless. I will add a big check that crop was called in the beginning of the method

seismicpro/src/seismic_batch.py
anton-br: If the others agree, let's leave it at that.

seismicpro/src/seismic_batch.py
```diff
@@ -1485,7 +1570,93 @@ def crop(self, index, src, coords, shape, dst=None):
...
+
+        coords_x = np.arange(0, len_x, x)
+        if len_x % x != 0 and drop_last:
+            coords_x = coords_x[:-1]
+
+        coords_y = np.arange(0, len_y, y)
+        if len_y % y != 0 and drop_last:
+            coords_y = coords_y[:-1]
+
+        getattr(self, dst)[pos] = list(product(coords_x, coords_y))
```
dpodvyaznikov, line 1659: Just as an idea to discuss: maybe we can store those not in separate component, but in the `meta`?

seismicpro/src/seismic_batch.py
```diff
@@ -1471,8 +1556,8 @@ def crop(self, index, src, coords, shape, dst=None):
             crop(src=['raw', 'mask], dst=['raw_crop', 'mask_crop], shape=(100, 256),
                 coords=P(R('uniform', size=(N_RANDOM_CROPS, 2)))).next_batch(2)
         """
-        if not isinstance(self.index, FieldIndex):
-            raise NotImplementedError("Index must be FieldIndex, not {}".format(type(self.index)))
+        if isinstance(self.index, SegyFilesIndex):
```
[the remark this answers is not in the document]
a-arefina: Yes, I've checked that this action works with `TraceIndex`

seismicpro/src/seismic_batch.py
a-arefina: It is really unclear how to use some arbitrary coordinates from parameters. I'll just fix the docstring

seismicpro/src/seismic_batch.py
a-arefina: I thought about that, but I don't like that in this case we'd have to create the component for crops to fill its meta before actual cropping

seismicpro/src/seismic_batch.py
```diff
@@ -1485,7 +1570,93 @@ def crop(self, index, src, coords, shape, dst=None):
         else:
             xy = np.array(coords)
 
-        getattr(self, dst)[pos] = self._crop(field, xy, shape)
+        getattr(self, dst)[pos] = self._crop(field, xy, shape, pad_zeros)
+
+        self.meta[dst].setdefault('source', src)
```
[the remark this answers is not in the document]
a-arefina: Good question. 

What about special `_init_for_crop` to use in @inbatch_parallel(init='_init_crop'), that inits component, and prepares meta for the current batch before parallelization? we could also clean up meta from the previous batch or action if needed. That will make reuse of components safer. The only issue is that such _init_for_action methods should be used with concrete actions.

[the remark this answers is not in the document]
dpodvyaznikov: To wrap up our phone talk: this can be achieved by making an action to deal with the meta, then call a method with `inbatch_parallel`, as it is done in `sort`.

seismicpro/src/seismic_batch.py
```diff
@@ -1075,6 +1077,74 @@ def seismic_plot(self, src, index, wiggle=False, xlim=None, ylim=None, std=1, #
...
+        title : str
+            Plot title.
+        kwargs : dict
+            Additional keyword arguments for plot.
+
+        Returns
+        -------
+        Multi-column subplots.
+        """
+        pos = self.get_pos(None, 'indices', index)
```
dpodvyaznikov, line 1121: You wanted to add a check that `crop` was called previously. Have you decided not to do so?

a-arefina: done

a-arefina:
Added tests for crop-related actions. Test data to be used is introduced in #26 
No changes are made to github's status.yml, because we need to decide whether test data will be stored in the git repo or using dvc

seismicpro/src/seismic_batch.py
```diff
@@ -1504,13 +1506,34 @@ def _crop(self, image, coords, shape, pad_zeros):
                     raise ValueError('Coordinates', (x, y), 'exceed feasible region of seismogram with shape',
                                      image.shape, ', with crop shape', shape, 'but pad_zeros is False')
                 res[i] = image[x:x+shape[0], y:y+shape[1]]
-                
+
         return res
 
+
+    @inbatch_parallel(init='indices')
+    def __crop(self, index, src, coords, shape, pad_zeros, dst=None):
```
anton-br, line 1514: We've never used two underscores in function name, what is your purpose to use it here?

seismicpro/src/seismic_batch.py
```diff
@@ -1468,7 +1468,9 @@ def equalize(self, index, src, dst, params, survey_id_col=None):
         getattr(self, dst)[pos] = equalized_field
         return self
 
-    def _crop(self, image, coords, shape, pad_zeros):
+
+    @staticmethod
+    def __crop_from_image(image, coords, shape, pad_zeros):
```
anton-br, line 1473: Why do you move this fragment to different function? Are there any ways to use it not in `__crop` function?

seismicpro/src/tests/crop_test.py
```diff
@@ -0,0 +1,115 @@
...
+
+PARAMS_LIST = [(*it_bs, cs) for it_bs in PARAMS for cs in PARAMS[it_bs]]
+
+
+class TestCropAssemble:
+    """ Check croppping and assembling """
+
+    @pytest.mark.parametrize('single_iteration', [True, pytest.param(False, marks=pytest.mark.slow)])
+    @pytest.mark.parametrize('index_type,batch_size,crop_shape', PARAMS_LIST)
+    def test_crop_assemble(self, index_type, batch_size, crop_shape, single_iteration):
```
anton-br, line 30: How we usually document tests? I think it should be document as always because further tests will be based on this example :)

seismicpro/src/tests/crop_test.py
```diff
@@ -0,0 +1,115 @@
...
+
+        index = index_type(path=PATH, name='raw')
+
+        ppl = (
+            SeismicDataset(index).p
+            .init_variable('raw', default=[])
+            .init_variable('crops', default=[])
+            .load(components='raw', fmt='segy', tslice=slice(2000))
+            .update(V('raw', 'a'), B('raw'))
+            .crop(src='raw', dst='crops', coords=[(0, 0), (0, 0.5)], shape=(1, 1), pad_zeros=True)
```
anton-br, line 100: I don't get how will to works here. Do I understand correctly that there are only 2 elements in the test dataset?

seismicpro/src/seismic_batch.py
a-arefina: This is a private method

seismicpro/src/tests/crop_test.py
a-arefina: From every element in every batch we are getting 2 crops of shape 1x1, one at coordinates (0, 0), the other at coordinates (0, (trace_len - 1)/2 ), that is (0, 999)

seismicpro/src/tests/crop_test.py
a-arefina: I've written short descriptions to each test, telling what they are doing. 
Describing parameters is useless, imo, because tests are called by a test framework, and no one will have to pass those parameters manually

So what sort of documentation you think is missing here?

seismicpro/src/seismic_batch.py
anton-br: doesn't " _ " mean that the method is private?

seismicpro/src/tests/crop_test.py
anton-br: I mean our usual format of docs with `Parameters` and other staff. Your answer suits me perfectly.

seismicpro/src/seismic_batch.py
a-arefina: Single leading underscore is just a naming convention: see https://docs.python.org/3.5/tutorial/classes.html#private-variables

a-arefina:
This waits for #26 with test data

seismicpro/src/seismic_batch.py
```diff
@@ -1471,21 +1558,105 @@ def crop(self, index, src, coords, shape, dst=None):
...
+    def assemble_crops(self, index, src, dst):
+        """
+        Assembles crops from `src` into a single seismogram
+
+        Parameters
+        ----------
+        src : str
+            component with crops
+        dst : str
+            component to put the result to. If None, `dst` == `src`
```
nikita-klsh, line 1585: The second sentence can be removed, its more about decorator, not current action

seismicpro/src/seismic_batch.py
```diff
@@ -1471,21 +1558,105 @@ def crop(self, index, src, coords, shape, dst=None):
...
                 coords=P(R('uniform', size=(N_RANDOM_CROPS, 2)))).next_batch(2)
         """
-        if not isinstance(self.index, FieldIndex):
-            raise NotImplementedError("Index must be FieldIndex, not {}".format(type(self.index)))
+        if isinstance(self.index, SegyFilesIndex):
+            raise NotImplementedError("Index can't be SegyFilesIndex")
+
+        self._init_component(dst=dst)
+        self.meta[dst]['crop_coords'] = {}
+        self.meta[dst]['source'] = src
```
nikita-klsh, line 1566: do we even need this key in meta? if so consider more meaningful name, like `source for crops`

seismicpro/src/seismic_batch.py
```diff
@@ -1471,21 +1558,105 @@ def crop(self, index, src, coords, shape, dst=None):
...
+            component with crops
+        dst : str
+            component to put the result to. If None, `dst` == `src`
+        """
+
+        if isinstance(self.index, SegyFilesIndex):
+            raise NotImplementedError("Index can't be SegyFilesIndex")
+
+        if 'crop_coords' not in self.meta[src]:
+            raise ValueError("{} component doesn't contain crops!".format(src))
```
nikita-klsh, line 1592: the following intuitive pipeline for inference
```
.crop(src='raw, dst='crops, coords=coords')
.predict(B('crops'), save_to=B('predictions').
.assemble_crops(src='predictions', dst='ass_predictions')
```
Wouldn't work since component `predictions` does not have key `crop_coords` in meta and meta at all.
Why we do not passing coords explicitly to the action `.assemble_crops(src='predictions', dst='ass_predictions', coords=coords)`. Please remind the motivation for storing the coords in the meta

seismicpro/src/seismic_batch.py
```diff
@@ -1471,21 +1558,105 @@ def crop(self, index, src, coords, shape, dst=None):
...
+
+        pos = self.get_pos(None, None, index)
+        crops = getattr(self, src)[pos]
+        coords = self.meta[src]['crop_coords'][index]
+
+        res_x = self.index.tracecounts[pos]
+        res_y = len(self.meta[self.meta[src]['source']]['samples'])
+
+        res = np.zeros((res_x, res_y))
```
nikita-klsh, line 1601: Should we consider the parameter `fill_value` to fill the missing values? `0` might have a sense for amplitudes, but not for probabilities

seismicpro/src/seismic_batch.py
```diff
@@ -1471,21 +1558,105 @@ def crop(self, index, src, coords, shape, dst=None):
...
+            x, y = crop_coords
+            len_x, len_y = crop.shape
+
+            x1 = min(x+len_x, res_x)
+            y1 = min(y+len_y, res_y)
+
+            res[x:x1, y:y1] = crop[:x1-x, :y1-y]
+            crop_counts[x:x1, y:y1] += 1
+
+        crop_counts[crop_counts == 0] = 1
```
nikita-klsh, line 1614: Filling the incomplete crops with some values is implicit, consider adding the flag that allows it.

Also consider the parameter `fill_value` to fill the missing values. `0` might have a sense for amplitudes, but not for probabilities

seismicpro/src/seismic_batch.py
```diff
@@ -1471,21 +1558,105 @@ def crop(self, index, src, coords, shape, dst=None):
...
+
+        getattr(self, dst)[pos] = res / crop_counts
+
+        return self
+
+
+    @action
```
nikita-klsh, line 1621: Placing this action at the beginning instead of the end seems more logical

seismicpro/src/seismic_batch.py
```diff
@@ -1471,21 +1558,105 @@ def crop(self, index, src, coords, shape, dst=None):
...
+    @inbatch_parallel(init='_init_component')
+    @apply_to_each_component
+    def make_grid_for_crops(self, index, src, dst, shape, drop_last=True):
```
nikita-klsh, line 1624: Should we consider adding `strides` parameter like its done in [seismiqb](https://github.com/gazprom-neft/seismiqb/blob/c0b309e5e17a79a3e7e6b2cafe1bee675575bcb2/seismiqb/src/cubeset.py#L374)? It might be useful to assemble intersecting crops

seismicpro/src/seismic_batch.py
```diff
@@ -1471,21 +1558,105 @@ def crop(self, index, src, coords, shape, dst=None):
...
         field = getattr(self, src)[pos]
 
-        if all(((0 < x < 1) and (0 < y < 1)) for x, y in coords):
-            feasible_region = np.array(field.shape) - shape
-            xy = (feasible_region * coords).astype(int)
-            if np.any(xy < 0):
-                raise ValueError("`shape` is larger than one of seismogram's dimensions")
-        else:
-            xy = np.array(coords)
+        len_x, len_y = field.shape
```
nikita-klsh, line 1645: it was `rex_x` and `res_y` in previous action

seismicpro/src/seismic_batch.py
```diff
@@ -1394,36 +1468,48 @@ def equalize(self, index, src, dst, params, survey_id_col=None):
...
-            res[i] = image[x:x+shape[0], y:y+shape[1]]
-        return res
+        pos = self.get_pos(None, src, index)
+        arr = getattr(self, src)[pos]
+
+        if all(((0 <= x < 1) and (0 <= y < 1)) for x, y in coords):
+            feasible_region = np.array(arr.shape) - shape
+            xy = (feasible_region * coords).astype(int)
+            if np.any(xy < 0):
+                raise ValueError("`shape` is larger than one of seismogram's dimensions")
```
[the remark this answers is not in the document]
a-arefina: The logic in float coordinates branch is as follows: crop shape is subtracted from image shape, and resulting crop coordinates are calculated using only that feasible region. So if we have image of shape (10, 10) and we want a crop of shape (4, 4) at relative coordinates (0.3, 0.5), then actual crop coordinates will be (2, 3). I did not change anything here from the previous version of crop action
Using this logic it `pad_zeros` is useless and is hard to decide what to do if the crops shape is larger that images shape.
I added a note on this in `pad_zeros` description in 'crop` docstring

seismicpro/models/unet_attention.py
```diff
@@ -19,12 +19,14 @@ def default_config(cls):
...
+    @classmethod
+    def initial_block(cls, inputs, name='initial_block', **kwargs):
+        _ = name, kwargs
         return inputs
 
-    def body(self, inputs, *args, **kwargs):
-        _ = args
+    @classmethod
+    def body(cls, inputs, name='body', **kwargs):
+        _ = name
```
roman-kh, line 29: Is it good to ignore the name? You can use it to as a scope which wraps both main and attention or as name for main branch. Don't ignore.

seismicpro/models/unet_attention.py
```diff
@@ -19,12 +19,14 @@ def default_config(cls):
 
         return config
 
-    def initial_block(self, inputs, *args, **kwargs):
-        _ = args, kwargs
+    @classmethod
+    def initial_block(cls, inputs, name='initial_block', **kwargs):
```
roman-kh, line 23: TFMode.initial_block returns inputs when layout is empty. So there seems no point in redefining it.

seismicpro/models/unet_attention.py
```diff
@@ -34,8 +36,8 @@ def body(self, inputs, *args, **kwargs):
         att = super().body(raw, name='attention', **{**kwargs, **attn_config}) # pylint: disable=not-a-mapping
         return main, att, raw, offset
 
-    def head(self, inputs, *args, **kwargs):
-        _ = args, kwargs
+    def head(self, inputs, targets, name='head', **kwargs):
+        _ = targets, name, kwargs
```
roman-kh, line 40: Same - so not ignore name. Wrap everything into a context with a given name.

seismicpro/models/unet_attention.py
a-arefina: Deleting `initial_block` redefinition somehow breaks mask detection. I'm still investigating it. 

I don't think it's blocking this PR, because this code is already in master, and it was changed only to satisfy pylint

PR #27, opened 2020-09-01
fix x_lim, y_lim

a-arefina:
When several arrays with different shapes are passed to `seismic_plot` and `xlim, ylim` are not set, resulting plots have the same shape, which is the shape of the array that is listed first

This PR fixes this error


dpodvyaznikov (approved):
Please, merge master into this branch to fix pylint and it's done

PR #29, opened 2020-09-11
latest batchflow

a-arefina:
`apply_transform` from the previous version of bf is now renamed to `apply_parallel`

`apply_transform_all` is removed - one can use `Pipeline.call` method, or add_namespace + directly calling the needed function


a-arefina:
У, класс! Тесты работают! Мерджить пока нельзя - надо посмотреть, почему они сломались

seismicpro/src/seismic_batch.py
```diff
@@ -460,7 +460,7 @@ def _dump_segy(self, src, path, split=True):
     @inbatch_parallel(init="indices", target="threads")
     def _dump_split_segy(self, index, src, path):
         """Dump data to segy files."""
-        pos = self.get_pos(None, src, index)
+        pos = self.index.get_pos(ndex)
```
SergeyTsimfer, line 463: Typo

SergeyTsimfer (commented):
Have you thought about implementing your own `Batch.get` method, that would encapsulate the need to retrieve `pos` over and over again?

seismicpro/src/seismic_dataset.py
```diff
@@ -136,7 +136,7 @@ def find_equalization_params(self, batch, component, survey_id_col, sample_size=
             check_unique_fieldrecord_across_surveys(surveys_by_fieldrecord, idx)
             survey = surveys_by_fieldrecord[0]
 
-            pos = batch.get_pos(None, component, idx)
+            pos = batch.get_pos(idx)
```
nikita-klsh, line 139: `batch.index.get_pos(idx)`?

a-arefina:
[НОвый БФ](https://trello.com/c/h8s0iM14/53-%D0%BD%D0%BE%D0%B2%D1%8B%D0%B9-%D0%B1%D1%84)

seismicpro/src/file_utils.py
```diff
@@ -43,32 +45,49 @@ def write_segy_file(data, df, samples, path, sorting=None, segy_format=1):
...
+    tracecounts = 0
+    samples = None
+
+    for fname in files_list:
+        with segyio.open(fname, strict=False) as f:
+            if samples is None:
+                samples = f.samples
+            else:
+                if np.all(samples != f.samples):
+                    raise ValueError("Inconsistent samples in files!" +
```
nikita-klsh, line 78: Why are we checking only the equivalence of this attribute, when there are more of them, like `sorting`.
I suggest removing at all. Or consider a more fair check.

seismicpro/src/file_utils.py
```diff
@@ -43,32 +45,49 @@ def write_segy_file(data, df, samples, path, sorting=None, segy_format=1):
...
+    if len(paths) == 0:
+        raise ValueError("`paths` cannot be empty.")
+
+    files_list = [f for path in paths for f in glob.iglob(path, recursive=True)]
+
+    tracecounts = 0
+    samples = None
+
+    for fname in files_list:
+        with segyio.open(fname, strict=False) as f:
```
nikita-klsh, line 73: Setting the flag `ignore_geometry=True` in context manager every time we open the file will significantly speed up the opening.

seismicpro/src/file_utils.py
a-arefina: Sorting can be changed afterwards, to be the same for all items, but inconsistent samples cannot be fixed. What other attributes do you think should be checked?

seismicpro/src/file_utils.py
a-arefina: Well, this is what is written in segyio docs: "If ignore_geometry=True, segyio will not try to build iline/xline or other geometry related structures, which leads to faster opens. This is essentially the same as using strict=False on a file that has no geometry."

I thought that this is exactly our case

Any thoughts from @dpodvyaznikov?

nikita-klsh: The behavior will be the same. 
However, for some reason in case `strict=False, ignore_geometry=False` geometry is being infered anyway, see the [source code](https://github.com/equinor/segyio/blob/366caffdc1d99202d474eaeac1bc0ada09838a57/python/segyio/open.py#L184).

seismicpro/src/tests/dump_merge_test.py
```diff
@@ -0,0 +1,116 @@
...
+
+    ppl.run(batch_size=1, n_epochs=1, drop_last=False, shuffle=False, bar=False)
+
+    res = np.stack(ppl.get_variable('res'))
+
+    assert np.all(res)
+
+
+@pytest.mark.parametrize('index_type', [FieldIndex, TraceIndex])
+def test_dump_split_merge(index_type, tmp_path):
```
dpodvyaznikov, line 62: Where does `tmp_path` value come from?

seismicpro/src/tests/dump_merge_test.py
```diff
@@ -0,0 +1,116 @@
...
+
+from seismicpro.batchflow import V, B, L, I
+from seismicpro.src import SeismicDataset, FieldIndex, TraceIndex, merge_segy_files
+
+PATH = os.path.join(os.path.dirname(os.path.realpath(__file__)), '../../../datasets/demo_data/teapot_dome_10.sgy')
+# 10 field records with shapes (912, 2049) or (966, 2049)
+
+
+def compare_files(path_1, path_2, compare_all):
+    """ Checks that the content of a given SGY-files is a subset of (or is equal to)
```
dpodvyaznikov, line 17: IMO this function needs more detailed explanation.

seismicpro/src/file_utils.py
```diff
@@ -43,41 +45,59 @@ def write_segy_file(data, df, samples, path, sorting=None, segy_format=1):
         for i, x in enumerate(file.header[:]):
             x.update(meta[i])
 
-def merge_segy_files(output_path, bar=True, **kwargs):
+def merge_segy_files(paths, output_path, bar=True):
```
dpodvyaznikov, line 48: We used `path` instead of `paths` before (as in `FilesIndex`).

I'd suggest either to change this argument to `path`, or update our notebooks:
tutorials/2.Batch.ipynb
tutorials/4.Models.ipynb
tutorials/5. Preprocessing.ipynb
models/Ground-roll_attenuation/Attention_model/model_description.ipynb
models/Ground-roll_attenuation/Unet_1D_model/model_description.ipynb

seismicpro/src/tests/dump_merge_test.py
a-arefina: It's pytest fixture - a temporary path that is created on each test run

dpodvyaznikov: Nice

nikita-klsh:
Please do the same transform `batch.get_pos -> batch.index.get_pos` in `seismic_dataset.py` methods

a-arefina:
> Please do the same transform `batch.get_pos -> batch.index.get_pos` in `seismic_dataset.py` methods

There is a single `get_pos` in `seismic_dataset.py` - and it's already fixed

dpodvyaznikov (approved):

PR #33, opened 2020-09-24
Metrics

anton-br:
This pull request is adding a oppotunity to cunstruct a metrics map. It allows to use `gather_metrics` from batchblow.


seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,110 @@
...
+    std = nanstd
+    quantile = quantile
+    absquantile = absquantile
+
+class MetricsMap(Metrics):
+    """seismic metrics class"""
+
+    AVALIBALE_METRICS = [
+        'construct_map'
+    ]
```
a-arefina, line 47: where do you use this?

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,110 @@
...
+            bin_size = (bin_size, bin_size)
+
+        maps_list_transposed = np.array(self.maps_list).T
+
+        coords_x = np.array(maps_list_transposed[0], dtype=np.float32)
+        coords_y = np.array(maps_list_transposed[1], dtype=np.float32)
+        metrics = np.array(list(maps_list_transposed[2]), dtype=np.float32)
+
+        call_aggr_bins = getattr(NumbaNumpy, aggr_bins)
+        call_aggr_bins = call_aggr_bins(None, **aggr_bins_kwargs) if aggr_bins_kwargs is not None else call_aggr_bins
```
a-arefina, line 84: the choice expression returns call_aggr_bins result in one branch and call_aggr_bins itself in the other

did you want to use partial function?


I would move this selection inside NumbaNumpy. For example, all its class methods could return callables

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,110 @@
...
+        bin_size_x, bin_size_y = bin_size
+        range_x = np.arange(coords_x.min(), coords_x.max() + bin_size_x, bin_size_x)
+        range_y = np.arange(coords_y.min(), coords_y.max() + bin_size_y, bin_size_y)
+        metrics_map = np.full((len(range_y), len(range_x)), np.nan)
+
+        for i in prange(len(range_x)): #pylint: disable=not-an-iterable
+            for j in prange(len(range_y)): #pylint: disable=not-an-iterable
+                mask = ((coords_x - range_x[i] >= 0) & (coords_x - range_x[i] < bin_size_x) &
+                        (coords_y - range_y[j] >= 0) & (coords_y - range_y[j] < bin_size_y))
+                if mask.sum() > 0:
```
a-arefina, line 108: np.any ?

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,110 @@
...
+        if isinstance(bin_size, int):
+            bin_size = (bin_size, bin_size)
+
+        maps_list_transposed = np.array(self.maps_list).T
+
+        coords_x = np.array(maps_list_transposed[0], dtype=np.float32)
+        coords_y = np.array(maps_list_transposed[1], dtype=np.float32)
+        metrics = np.array(list(maps_list_transposed[2]), dtype=np.float32)
+
+        call_aggr_bins = getattr(NumbaNumpy, aggr_bins)
```
a-arefina, line 83: aggr_bins - what about renaming to aggr_bins_fn ?

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,110 @@
...
+        """njit map"""
+        bin_size_x, bin_size_y = bin_size
+        range_x = np.arange(coords_x.min(), coords_x.max() + bin_size_x, bin_size_x)
+        range_y = np.arange(coords_y.min(), coords_y.max() + bin_size_y, bin_size_y)
+        metrics_map = np.full((len(range_y), len(range_x)), np.nan)
+
+        for i in prange(len(range_x)): #pylint: disable=not-an-iterable
+            for j in prange(len(range_y)): #pylint: disable=not-an-iterable
+                mask = ((coords_x - range_x[i] >= 0) & (coords_x - range_x[i] < bin_size_x) &
+                        (coords_y - range_y[j] >= 0) & (coords_y - range_y[j] < bin_size_y))
```
a-arefina, line 107: I feel that there are more clear and readable ways to define mask (are slices ok here?), but I believe that you made all this complicated mask creation for the sake of efficiency.

tutorials/metrics_map_pipeline.ipynb
```diff
@@ -0,0 +1,135 @@
+{
```
a-arefina, line 1: <p>1) Please remove unuse imports like <code>sns</code></p><p>2) I couldn't find <code>SemblanceMetrics</code> . did you commit it?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gazprom-neft/SeismicPro/pull/33/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

a-arefina, line 1: <p>I'd like to have a more convenient way to define <code>coords</code> in gather_metrics</p><p> </p><p>BTW, did you check this with a larger batch size?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gazprom-neft/SeismicPro/pull/33/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

a-arefina, line 1: <p>I couldn't understand how <code>'map'</code>  is transformed to <code>construct_map</code>  , i.e. how do you add <code>construct_</code>  to the method name</p><p> </p><p>All in all, one would need a good explanatory notebook for this feature, so please add more comments</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gazprom-neft/SeismicPro/pull/33/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

a-arefina (commented):
I feel that tests will be super-helpful in writing a better structured code

seismicpro/src/plot_utils.py
```diff
@@ -530,3 +531,27 @@ def show_2d_heatmap(idf, figsize=None, save_to=None, dpi=300, **kwargs):
...
+        colors = ((0.0, 0.6, 0.0), (.66, 1, 0), (0.9, 0.0, 0.0))
+        cm = LinearSegmentedColormap.from_list(
+            'cm', colors)
+        cm.set_under('black')
+        cm.set_over('red')
+
+    fig = plt.figure(figsize=figsize)
+    img = plt.imshow(metrics_map, origin='lower', interpolation='nearest',
+                    vmin=vmin, vmax=vmax, cmap=cm, aspect='auto',
+                    extent=extent_coords, **kwargs)
```
nikita-klsh, line 550: feels like if we do not specify `extent` at all, it anyway does the same calculations under the hood

nikita-klsh:
Is such kind of padding expected?

```
x, y, m = np.random.randint(1, 101, (3, 10000)) # x and y take value 1 to 100
res = construct_metrics_map(x, y, m, (10, 10), NumbaNumpy.mean) # bin size = 10x10
plt.imshow(res)
```
image

anton-br:
> Is such kind of padding expected?
> 
> ```
> x, y, m = np.random.randint(1, 101, (3, 10000)) # x and y take value 1 to 100
> res = construct_metrics_map(x, y, m, (10, 10), NumbaNumpy.mean) # bin size = 10x10
> plt.imshow(res)
> ```
> 
> image

no, it isn't expected, thanks

seismicpro/src/seismic_metrics.py
```diff
@@ -78,33 +74,50 @@ def construct_map(self, bin_size=500, vmin=None, vmax=None, cm=None, title=None,
...
         coords_y = np.array(maps_list_transposed[1], dtype=np.float32)
-        metrics = np.array(list(maps_list_transposed[2]), dtype=np.float32)
-
-        call_aggr_bins = getattr(NumbaNumpy, aggr_bins)
-        call_aggr_bins = call_aggr_bins(None, **aggr_bins_kwargs) if aggr_bins_kwargs is not None else call_aggr_bins
-
-        metric_map = self.construct_metrics_map(coords_x=coords_x, coords_y=coords_y,
+        metrics = np.array(list(maps_list_transposed[2]))
+        if isinstance(metrics[0], (tuple, list, set, np.ndarray)):
+            if len(metrics[0].shape) > 1:
```
a-arefina, line 79: In this branch `metrics[0]` can be `tuple, list, set`, but none of those have `shape` attribute

seismicpro/src/seismic_metrics.py
```diff
@@ -78,33 +74,50 @@ def construct_map(self, bin_size=500, vmin=None, vmax=None, cm=None, title=None,
...
+                raise ValueError('Construct map does not work with 3d metrics yet.')
```
a-arefina, line 80: what about 2d?

seismicpro/src/seismic_metrics.py
```diff
@@ -78,33 +74,50 @@ def construct_map(self, bin_size=500, vmin=None, vmax=None, cm=None, title=None,
 
         coords_x = np.array(maps_list_transposed[0], dtype=np.float32)
```
a-arefina, line 75: It might be more convenient to use self.coords and seld.metrics without packing them into self.maps_list and then unpacking again

You can extend both of them in `self.extend`

seismicpro/src/seismic_metrics.py
```diff
@@ -78,33 +74,50 @@ def construct_map(self, bin_size=500, vmin=None, vmax=None, cm=None, title=None,
...
-        return metric_map
+                             title=title, figsize=figsize, save_to=save_to, pad=pad)
+        return metric_map, shapes
 
     @staticmethod
     @njit(parallel=True)
-    def construct_metrics_map(coords_x, coords_y, metrics, bin_size, aggr_bins):
+    def construct_metrics_map(coords_x, coords_y, metrics, bin_size, agg_bins_fn):
         """njit map"""
+        shapes = []
```
a-arefina, line 109: What about `shapes` being a (range_x, range_y) array? It might be mere convenient.

anton-br: this thing shouldn't be here, ill remove it completely because numba can't work with list

seismicpro/src/seismic_metrics.py
anton-br: thanks, i'll change it

seismicpro/src/seismic_metrics.py
anton-br: here, 2d means that for each coordinates we have an array with metrics. Because when you will calcualte metircs in bin, you will receive 2d array.
And yes, metrics works with 2d arrays now

seismicpro/src/plot_utils.py
anton-br: no it is not, because metrics map has indices from 0 to `metrics_map.shape` but we want to write our coordinates by axis

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,110 @@
...
+
+def quantile(array, q):
+    """ numba quantile. """
+    _ = array
+    return  njit(lambda array: np.quantile(array, q=q))
+
+def absquantile(array, q):
+    """ numba absquantile. """
+    _ = array
+    return njit(lambda array: _absquantile(array, q=q))
```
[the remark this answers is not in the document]
anton-br: yes, i do, because you can't simply pass arguments to numba function..

seismicpro/src/seismic_metrics.py
anton-br: here i want to create a callable that one can call without arguments. It means that in the case if `call_aggr_bins` has arguments, we will add them here and the `call_aggr_bins` should return another callable that can be used without arguments

tutorials/metrics_map_pipeline.ipynb
anton-br: i will delete this notebook complitly

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,123 @@
...
+
+@njit
+def _absquantile(array, q):
+    shifted_array = array - np.mean(array)
+    val = np.quantile(np.abs(shifted_array), q)
+    ind = np.argmin(np.abs(np.abs(shifted_array) - val))
+    return array[ind]
+
+class NumbaNumpy:
```
nikita-klsh, line 27: There is [jitclass](https://numba.pydata.org/numba-doc/dev/user/jitclass.html#jitclass) decorator that decorates each method of the class with `njit` decorator. It might simplify things here

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,122 @@
...
+            call_agg_bins = call_agg_bins(None, **agg_bins_kwargs) if agg_bins_kwargs is not None else call_agg_bins
+        elif callable(agg_bins_fn):
+            call_agg_bins = agg_bins_fn
+        else:
+            raise ValueError('agg_bins_fn should be whether str or callable, not {}'.format(type(agg_bins_fn)))
+
+        metric_map = self.construct_metrics_map(coords_x=coords_x, coords_y=coords_y,
+                                                metrics=metrics, bin_size=bin_size,
+                                                agg_bins_fn=call_agg_bins)
+        extent_coords = [coords_x.min(), coords_x.max(), coords_y.min(), coords_y.max()]
```
nikita-klsh, line 69: This list is only used in the plot function, so it can be created directly there

seismicpro/src/seismic_metrics.py
anton-br: unfortunatelly, it seems that you can't choose aggregation in bins (or i just missed it)

seismicpro/src/plot_utils.py
```diff
@@ -543,10 +543,12 @@ def plot_metrics_map(metrics_map, vmin=None, vmax=None, extent_coords=None, cm=N
         cm.set_under('black')
         cm.set_over('red')
 
+    origin = kwargs.get('origin', 'lower')
+    aspect = kwargs.get('aspect', 'auto')
```
a-arefina, line 547:
```suggestion
    origin = kwargs.pop('origin', 'lower')
    aspect = kwargs.pop('aspect', 'auto')
```

anton-br:
Consider to add `no-name-in-module` and `import-error` into pylintrc because i don't know how to fix these errors.

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,135 @@
...
+    ind = np.argmin(np.abs(np.abs(shifted_array) - val))
+    return array[ind]
+
+class NumbaNumpy:
+    """ Holder for jit-accelerated functions. """
+    #pylint: disable = unnecessary-lambda, undefined-variable
+    nanmin = njit(lambda array: np.nanmin(array))
+    nanmax = njit(lambda array: np.nanmax(array))
+    nanmean = njit(lambda array: np.nanmean(array))
+    nanstd = njit(lambda array: np.nanstd(array))
```
a-arefina, line 128:
```suggestion
    nanstd = njit(np.nanstd)
```
Can this be?

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,135 @@
...
+
+def quantile(array, q):
+    """ numba quantile. """
+    _ = array
+    return  njit(lambda array: np.quantile(array, q=q))
+
+def absquantile(array, q):
+    """ numba absquantile. """
+    _ = array
+    return njit(lambda array: _absquantile(array, q=q))
```
a-arefina, line 113:
```suggestion
    return njit(lambda arr: _absquantile(arr, q=q))
```

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,135 @@
...
+    nanmin = njit(lambda array: np.nanmin(array))
+    nanmax = njit(lambda array: np.nanmax(array))
+    nanmean = njit(lambda array: np.nanmean(array))
+    nanstd = njit(lambda array: np.nanstd(array))
+
+    min = nanmin
+    max = nanmax
+    mean = nanmean
+    std = nanstd
+    quantile = quantile
```
a-arefina, line 134:
```suggestion
    def quantile(q): return  njit(lambda array: np.quantile(array, q=q))
```

And don't pass `None` to `call_agg_bins` when creating it with kwargs

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,135 @@
...
+            metrics_storage.fill(np.nan)
+            mask = np.zeros_like(metrics_storage) + np.arange(np.max(metrics_shapes))
+            mask = mask < metrics_shapes
+            metrics_storage[mask] = np.concatenate(metrics)
+            metrics = metrics_storage.copy()
+        metrics = np.array(metrics, dtype=np.float32)
+
+        if isinstance(agg_bins_fn, str):
+            call_agg_bins = getattr(NumbaNumpy, agg_bins_fn)
+            call_agg_bins = call_agg_bins(None, **agg_bins_kwargs) if agg_bins_kwargs is not None else call_agg_bins
```
a-arefina, line 71:
```suggestion
            call_agg_bins = call_agg_bins(**agg_bins_kwargs) if agg_bins_kwargs else call_agg_bins
```
see comment to line 134
if `agg_bins_kwargs` is an empty dict, you shouldn't pass it to `call_agg_bins`, should you?

seismicpro/src/seismic_metrics.py
anton-br: no, it can't. This way you'll create a callable with full list of arguments from np.nanstd function while numba works only with the first one. I receive the error `CALL_FUNCTION_EX with **kwargs not supported`.

seismicpro/src/seismic_metrics.py
anton-br: thank for the tip

pylintrc
```diff
@@ -13,12 +13,13 @@ argument-rgx=([a-z][a-z0-9_]{1,30}|[a-z_])$ # snake_case + single letters
...
 disable=no-member, no-value-for-parameter, no-self-use, too-many-locals, too-few-public-methods,
-        too-many-public-methods, too-many-branches, unsubscriptable-object, redefined-variable-type, 
+        too-many-public-methods, too-many-branches, unsubscriptable-object, redefined-variable-type,
         too-many-star-expressions, duplicate-code, not-context-manager, too-many-lines, global-statement,
         locally-disabled, wrong-import-position, invalid-sequence-index, redundant-keyword-arg, bad-super-call,
         no-self-argument, redefined-builtin, arguments-differ, len-as-condition, keyword-arg-before-vararg,
         assignment-from-none, useless-return, useless-import-alias, unnecessary-pass, cyclic-import,
-        assignment-from-no-return, comparison-with-callable, unnecessary-lambda, no-method-argument, blacklisted-name
+        assignment-from-no-return, comparison-with-callable, unnecessary-lambda, no-method-argument, blacklisted-name,
+        no-name-in-module, import-error
```
nikita-klsh, line 22: Why not to do this locally in the file?

anton-br: Do we really need these varnings? why?

nikita-klsh: Then we block a bunch of checks for the whole repository. And if we push the really bad piece of code, pylint will skip it silently.

Since it is the false warning,  you can add `pylint: disable` locally. Then maybe in the next release  pylint  will fix the case and throws the message that `pylint: disable` here is unnecessary. Then we remove it.

anton-br: no i mean do we really need these two varnings in our repository

nikita-klsh: Why you do not want to detect the imports of unexisting modules?

anton-br: because you will receive the error message from python in this case

nikita-klsh: If you execute code before pushing :)

nikita-klsh:
Please add the base class to `__init__.py`

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,138 @@
...
+                raise ValueError('agg_bins_fn should be whether str or callable, not {}'.format(type(agg_bins_fn)))
+
+            # We need to avoid recompiling the numba function if aggregation function hasn't changed.
+            if self.call_name is None or self.call_name != call_name:
+                self.call_name = call_name
+                self.call_agg_bins = call_agg_bins
+
+            metric_map.append(self.construct_metrics_map(coords_x=coords_x, coords_y=coords_y,
+                                                         metrics=metrics, bin_size=bin_size,
+                                                         agg_bins_fn=self.call_agg_bins))
```
nikita-klsh, line 88: It's very nice that we now can pass list of metrics to `construct map`.

However, for each metric we recalculating point-bin distribution over and over again, which is the same across metrics, only the aggregation function differs. 

Could you move this part out of metrics loop?

anton-br:
i've tried to change our implemetation to https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.binned_statistic_2d.html. But scipy works much slower for huge data.
image

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,210 @@
...
+        for name, metrics in kwargs.items():
+            if isinstance(metrics, (int, float, bool, np.number)):
+                setattr(self, name, [metrics])
+            elif isinstance(metrics, (list, tuple, set, np.ndarray)):
+                setattr(self, name, [*metrics])
+            else:
+                raise ValueError("Wrong type of metrics have been given. "\
+                                 "Should be number or array but {} received.".format(type(metrics)))
+
+        self.metrics_names = list(kwargs.keys())
```
alexandrkuvaev, line 57: Here you are hoping for the same returned metrics names order in `items` and `keys`, while generally it's not true.

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,210 @@
...
+    ----
+    1. All keys from kwargs become class attributes.
+    """
+
+    def __init__(self, coords, **kwargs):
+        super().__init__()
+        if not kwargs:
+            raise ValueError('Add at least one metrics.')
+
+        # Create attributes with metrics.
```
alexandrkuvaev, line 47: What if we iterate over a dataset with a batch size more than 1 and want to pass an array for each pair of coordinates?

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,210 @@
...
+            if isinstance(metrics, (int, float, bool, np.number)):
+                setattr(self, name, [metrics])
+            elif isinstance(metrics, (list, tuple, set, np.ndarray)):
+                setattr(self, name, [*metrics])
+            else:
+                raise ValueError("Wrong type of metrics have been given. "\
+                                 "Should be number or array but {} received.".format(type(metrics)))
+
+        self.metrics_names = list(kwargs.keys())
+        self.coords = list(coords)
```
alexandrkuvaev, line 58: Validate, that lengths of metrics match the length of `coords`. Better do it in the above loop.

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,210 @@
...
+    coords : array-like with length 2
+        Array with coordinates for X and Y axes.
+    kwargs : dict
+        All given kwargs are considered quality values. Key value - any name, while value must be
+        whether number or 1-dimensional array.
+
+    Attributes
+    ----------
+    metrics_names : array-like
+        names of existing metrics.
```
alexandrkuvaev, line 27: existing?

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,210 @@
...
+            else:
+                raise ValueError("Wrong type of metrics have been given. "\
+                                 "Should be number or array but {} received.".format(type(metrics)))
+
+        self.metrics_names = list(kwargs.keys())
+        self.coords = list(coords)
+        self.aggr_func = None
+        self.call_name = None
+
+        self._agg_fn_dict = {'mean': np.nanmean,
```
alexandrkuvaev, line 62: Unify `agg` and `aggr`, I like `agg` more. Moreover, this dict is unused and should be deleted.

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,210 @@
...
+                             'max': np.nanmax,
+                             'min': np.nanmin}
+
+    def extend(self, metrics):
+        """Extend coordinates and metrics to global container."""
+        # Extend all attributes with given metrics values.
+        for name in self.metrics_names:
+            metrics_attr = getattr(self, name)
+            metrics_attr.extend(getattr(metrics, name))
+            setattr(self, name, metrics_attr)
```
alexandrkuvaev, line 72: Why not one-line it as you did with `coords`?

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,210 @@
...
+        """ All optained coordinates, are splitted into bins of the specified `bin_size`. Resulted map
+        is an array in which every value reflects the metric's value in one current bin. If there are no values
+        are included in the bin, it's value is None. Otherwise, the value of this bin is calculated based
+        on the aggregation function `agg_func`.
+
+        Each value in resulted map represent aggregated value of metrics for coordinates belongs to current bin.
+
+        Parameters
+        ----------
+        metrics_name : str
```
alexandrkuvaev, line 87: `metric_name` then.

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,210 @@
...
+        for name in self.metrics_names:
+            metrics_attr = getattr(self, name)
+            metrics_attr.extend(getattr(metrics, name))
+            setattr(self, name, metrics_attr)
+
+        self.coords.extend(metrics.coords)
+
+    def construct_map(self, metrics_name, bin_size=500, aggr_func='mean',
+                      agg_func_kwargs=None, plot=True, **plot_kwargs):
+        """ All optained coordinates, are splitted into bins of the specified `bin_size`. Resulted map
```
alexandrkuvaev, line 78: Grammar! optained! splitted!

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,210 @@
...
+        is an array in which every value reflects the metric's value in one current bin. If there are no values
+        are included in the bin, it's value is None. Otherwise, the value of this bin is calculated based
```
alexandrkuvaev, line 80: Please, pass all your docstrings through Grammarly.

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,210 @@
...
+        on the aggregation function `agg_func`.
```
alexandrkuvaev, line 81: agg or aggr? Unify everywhere.

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,210 @@
...
+            will be divided into bins with size `bin_size`.
+            If int, the bin size will be same for X and Y dimensions.
+        aggr_func : str or callable, optional, default 'mean'
+            Function to aggregate metrics values in one bin.
+            If str, should be attribue's name from meth:`NumbaNumpy`.
+            if callable, it will be applied for every bin independently.
+        agg_func_kwargs : dict, optional
+            Kwargs that will be applied to aggr_func before evaluating.
+        plot : bool, optional, default True
+            If True, metrics will be plotted
```
alexandrkuvaev, line 100: If `False`?

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,210 @@
...
+        **plot_kwargs : dict
+            Kwargs that are passed directly to plotter, see :func:`.plot_utils.plot_metrics_map`.
+            (allowed arguments: cm, title, figsize, save_to, dpi, pad, font_size, x_ticks, y_ticks
+                               and kwargs to :func:`matplotlib.pyplot.imshow`.)
+        """
```
alexandrkuvaev, line 105: No `Returns` section.

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,210 @@
...
+            If str, should be attribue's name from meth:`NumbaNumpy`.
+            if callable, it will be applied for every bin independently.
+        agg_func_kwargs : dict, optional
+            Kwargs that will be applied to aggr_func before evaluating.
+        plot : bool, optional, default True
+            If True, metrics will be plotted
+        **plot_kwargs : dict
+            Kwargs that are passed directly to plotter, see :func:`.plot_utils.plot_metrics_map`.
+            (allowed arguments: cm, title, figsize, save_to, dpi, pad, font_size, x_ticks, y_ticks
+                               and kwargs to :func:`matplotlib.pyplot.imshow`.)
```
alexandrkuvaev, line 104: Delete.

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,210 @@
...
+        """
+        metrics = np.array(list(getattr(self, metrics_name)))
```
alexandrkuvaev, line 106: But metric value is already stored as a list, isn't it?

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,210 @@
...
+
+        if not len(metrics):
+            raise ValueError('Given metrics is empty.')
```
alexandrkuvaev, line 109: Remove as you add this check in `__init__`.

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,210 @@
...
+
+        # if metrics has an array for one coordinate, we repeat the coordinate value
+        # and expand the metric values into a one-dimensional array.
+        if isinstance(metrics[0], (list, tuple, set, np.ndarray)):
+            len_of_copy = [len(metrics_array) for metrics_array in metrics]
```
alexandrkuvaev, line 114: You'll never have a list of lists as a metric value due to your wrong `__init__` and `extend` implementation.

Currently, metrics passed as a list should not work correctly: after `extend` you will obtain a single list with all metric values for all the coordinates specified.

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,210 @@
...
+
+        if len(metrics) != len(coords):
+            raise ValueError("The length of given metrics is not corresponds with length of the coordinates.\
+                              Check the metrics array, it is souldn't have a nested structure.")
+
+        coords_x = np.array(coords[:, 0], dtype=np.int32)
+        coords_y = np.array(coords[:, 1], dtype=np.int32)
+        metrics = np.array(metrics, dtype=np.float32)
+
+        if isinstance(bin_size, int):
```
alexandrkuvaev, line 128: If it's float?

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,210 @@
...
+            aggr_func = aggr_func(**agg_func_kwargs) if agg_func_kwargs else aggr_func
+        elif callable(aggr_func):
+            call_name = aggr_func.__name__
+        else:
+            raise ValueError('aggr_func should be whether str or callable, not {}'.format(type(aggr_func)))
+
+        # We need to avoid recompiling the numba function if aggregation function hasn't changed.
+        if self.call_name is None or self.call_name != call_name:
+            self.call_name = call_name
+            self.aggr_func = aggr_func
```
alexandrkuvaev, line 143: Don't get these lines. What if you first call this method with `mean` aggregation, then `max` and then `mean` again?

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,210 @@
...
+
+        metric_map = self.construct_metrics_map(coords_x=coords_x, coords_y=coords_y,
+                                                metrics=metrics, bin_size=bin_size,
+                                                aggr_func=self.aggr_func)
+
+        if plot:
+            extent = [coords_x.min(), coords_x.max(), coords_y.min(), coords_y.max()]
+            # Avoid the situation when we have only one coordinate for x or y dimension.
```
alexandrkuvaev, line 151: What's wrong with this case?

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,210 @@
...
+            extent[1] += 1 if extent[0] - extent[1] == 0 else 0
+            extent[3] += 1 if extent[2] - extent[3] == 0 else 0
+            plot_metrics_map(metrics_map=metric_map, extent=extent, **plot_kwargs)
```
alexandrkuvaev, line 154: Unify metric and metrics.

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,210 @@
...
+        metrics = np.array(metrics, dtype=np.float32)
+
+        if isinstance(bin_size, int):
+            bin_size = (bin_size, bin_size)
+
+        if isinstance(aggr_func, str):
+            call_name = aggr_func
+            aggr_func = getattr(NumbaNumpy, aggr_func)
+            aggr_func = aggr_func(**agg_func_kwargs) if agg_func_kwargs else aggr_func
+        elif callable(aggr_func):
```
alexandrkuvaev, line 135: Why no use of `agg_func_kwargs` in this case?

seismicpro/src/seismic_metrics.py
alexandrkuvaev, line 80: Not `None` but `np.nan`

seismicpro/src/seismic_metrics.py
alexandrkuvaev: yep

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,210 @@
...
+        for i in prange(len(range_x)): #pylint: disable=not-an-iterable
+            for j in prange(len(range_y)): #pylint: disable=not-an-iterable
+                mask = ((coords_x - range_x[i] >= 0) & (coords_x - range_x[i] < bin_size_x) &
+                        (coords_y - range_y[j] >= 0) & (coords_y - range_y[j] < bin_size_y))
+                if mask.sum() > 0:
+                    metrics_map[j, i] = aggr_func(metrics[mask])
+        return metrics_map
+
+
+class NumbaNumpy:
```
alexandrkuvaev, line 188: I'll think of what we can do to avoid this class, looks scary.

alexandrkuvaev (changes requested):

seismicpro/src/seismic_metrics.py
anton-br: agree with first part. This dict is used to aggregate resulted metrucs map, see https://github.com/analysiscenter/batchflow/blob/master/batchflow/models/metrics/base.py.

seismicpro/src/seismic_metrics.py
anton-br: i don't care about the names order here. I used metrics_names only to `extend`  metrics' values.

seismicpro/src/seismic_metrics.py
anton-br: Because you have to update all metrics from `metrics_names`. I don't know how to do it in one-line. Please help me with this line if you know how to do it.

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,238 @@
...
+    """ Class for metrics aggregation and plotting. This class aims to accumulate
+    coordinates and metrics values for current coordinates. Therefore, all calculations
+    must be performed outside of the class.
+
+    Parameters
+    ----------
+    coords : array-like with length 2
+        Array with coordinates for X and Y axes.
+    kwargs : dict
+        All given kwargs are considered quality values. The key value is any name,
```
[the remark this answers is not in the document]
anton-br: I rewrote this block a bit, take a look.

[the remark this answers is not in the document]
dpodvyaznikov: Way better, imo :+1:

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,229 @@
...
+        """Calculation of metrics map.
+
+        Parameters
+        ----------
+        coords_x : array-like
+            Coordinates for X axis.
+        coords_x : array-like
+            Coordinates for Y axis.
+        metrics : array-like
+            Quality values.
```
alexandrkuvaev, line 208: Let's replace quality with something more concise

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,229 @@
...
+        bin_size : tuple with length 2
+            The size of bin by X and Y axes.
+        arrg_func : numba callable
+            Function to aggregate metrics values in one bin.
+
+        Returns
+        -------
+            : two-dimensional array
```
alexandrkuvaev, line 216: Call it some way. Or is it a `SeismicPro` docsting convention?

seismicpro/src/utils.py
```diff
@@ -792,3 +793,35 @@ def transform_to_fixed_width_columns(path, path_save=None, n_spaces=8, max_len=(
             if path_save:
                 return
             shutil.copyfile(write_file.name, path)
+
+def create_args(call, skip_first=False, **kwargs):
```
alexandrkuvaev, line 797: Maybe move it to `seismic_metrics.py`, now it's used only there and, hopefully, so be it.

seismicpro/src/utils.py
```diff
@@ -792,3 +793,35 @@ def transform_to_fixed_width_columns(path, path_save=None, n_spaces=8, max_len=(
...
+    """ Constructing tuple with positional arguments to callable ``call`` based on
+    ``kwargs`` and ``call``'s defaults.
+
+    Parameters
+    ----------
+    call : callable
+        Function to create positional arguments for.
+    skip_first : bool
+        If True, created tuple won't contain value for frist argument.
```
alexandrkuvaev, line 806:
```suggestion
        If True, created tuple won't contain value for the first argument.
```

seismicpro/src/utils.py
```diff
@@ -792,3 +793,35 @@ def transform_to_fixed_width_columns(path, path_save=None, n_spaces=8, max_len=(
...
+        if False, created tuple will contain value for first argument.
```
alexandrkuvaev, line 807:
```suggestion
        if False, created tuple will contain value for the first argument.
```

seismicpro/src/utils.py
```diff
@@ -792,3 +793,35 @@ def transform_to_fixed_width_columns(path, path_save=None, n_spaces=8, max_len=(
...
+
+def create_args(call, skip_first=False, **kwargs):
+    """ Constructing tuple with positional arguments to callable ``call`` based on
+    ``kwargs`` and ``call``'s defaults.
+
+    Parameters
+    ----------
+    call : callable
+        Function to create positional arguments for.
+    skip_first : bool
```
alexandrkuvaev, line 805: Not clear what this argument is responsible for

seismicpro/src/utils.py
```diff
@@ -792,3 +793,35 @@ def transform_to_fixed_width_columns(path, path_save=None, n_spaces=8, max_len=(
...
+    kwargs : dict
+        Keyword arguments to function ``call``.
+
+    Returns
+    -------
+        : tuple
+        Positional arguments to ``call``.
+    """
+    params = inspect.signature(call).parameters
+    args = list(np.zeros(len(params.keys()) - skip_first))
```
alexandrkuvaev, line 817:
```suggestion
    args = [kwargs.get(name, param.default) for name, param in params.items()]
```
Let's do this with the following checks empty parameters and `skip_first` flag. Should be much readable.

alexandrkuvaev (changes requested):

seismicpro/src/plot_utils.py
```diff
@@ -425,3 +426,106 @@ def show_2d_heatmap(idf, figsize=None, save_to=None, dpi=300, **kwargs):
...
+        The size of text.
+    ticks_labels_x : array-like, optional
+        Ticks labels for x axis. Passed directly to :func:`matplotlib.axes.Axes.set_xticklabels`.
+    ticks_labels_y : array-like, optional
+        Ticks labels for y axis. Passed directly to :func:`matplotlib.axes.Axes.set_yticklabels`.
+    x_ticks : int
+        The number of coordinates on the x-axis.
+    y_ticks : int
+        The number of coordinates on the y-axis.
+    save_to : str, optional
```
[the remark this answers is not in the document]
anton-br: others plotters have `save_to` argument, but we can discuss it

seismicpro/src/plot_utils.py
```diff
@@ -425,3 +426,110 @@ def show_2d_heatmap(idf, figsize=None, save_to=None, dpi=300, **kwargs):
...
+    ----------
+    ax : matplotlib axes
+        Axes to which coordinates are added.
+    img_shape : array with length 2
+        Shape of the image to add ticks to.
+    ticks_range_x : array-like with length 2, optional
+        Min and max value of labels on the x-axis.
+    ticks_range_y : array-like with length 2, optional
+        Min and max value of labels on the y-axis.
+    x_ticks : int, optional
```
dpodvyaznikov, line 512: Technically `x_ticks` and `y_ticks` are optional since they have a default - `None`, but practically this function would fail if one does not provide meaningful value for them.

seismicpro/src/plot_utils.py
```diff
@@ -425,3 +426,106 @@ def show_2d_heatmap(idf, figsize=None, save_to=None, dpi=300, **kwargs):
...
+    pad : bool
+        If true, edges of the figure will be padded with a thin white line.
+        otherwise, the figure will not change.
+    font_size : int
+        The size of text.
+    ticks_labels_x : array-like, optional
+        Ticks labels for x axis. Passed directly to :func:`matplotlib.axes.Axes.set_xticklabels`.
+    ticks_labels_y : array-like, optional
+        Ticks labels for y axis. Passed directly to :func:`matplotlib.axes.Axes.set_yticklabels`.
+    x_ticks : int
```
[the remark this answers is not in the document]
dpodvyaznikov: If parameter has a default value, it is an optional parameter.

dpodvyaznikov (approved):

seismicpro/src/seismic_metrics.py
```diff
@@ -0,0 +1,269 @@
...
+
+        if plot:
+            ticks_range_x = [coords_x.min(), coords_x.max()]
+            ticks_range_y = [coords_y.min(), coords_y.max()]
+            plot_metrics_map(metrics_map=metrics_map, ticks_range_x=ticks_range_x,
+                             ticks_range_y=ticks_range_y, **plot_kwargs)
+        return metrics_map
+
+    def _create_args(self, call, **kwargs):
+        """ Constructing tuple with positional arguments to callable ``call`` based on
```
alexandrkuvaev, line 213: Explicitly specify, that you omit the first arg

alexandrkuvaev (changes requested):

alexandrkuvaev (approved):

PR #37, opened 2020-11-09
Fix bug in meta

anton-br:
If you create component by hands it wound't have a meta at all. Old `copy_meta` always is tried to find new component in meta and fails if the component is not contained in the meta dict.

Moreover, this PR is added  empty meta for components that was created by `add_components` action.

### DIFF
--- a/seismicpro/src/seismic_batch.py
+++ b/seismicpro/src/seismic_batch.py
@@ -257,17 +257,79 @@ class SeismicBatch(Batch):
 
         for fr_comp, t_comp in zip(from_comp, to_comp):
             if fr_comp not in self.meta:
-                raise ValueError('{} does not exist.'.format(fr_comp))
+                raise ValueError('Meta of the component {} does not exist.'.format(fr_comp))
 
             if fr_comp == t_comp:
                 continue
 
-            if t_comp in self.meta:
+            if self.meta.get(t_comp):
                 warnings.warn("Meta of the component {} is not empty and".format(t_comp) + \
                               " will be replaced by the meta from the component {}.".format(fr_comp),
                               UserWarning)
+
             self.meta[t_comp] = self.meta[fr_comp].copy()
         return self
+
+    @action
+    def add_components(self, components, init=None):
+        """ Add new components
+
+        Parameters
+        ----------
+        components : str or list
+            new component names
+        init : array-like
+            initial component data
+
+        Raises
+        ------
+        ValueError
+            If a component or an attribute with the given name already exists
+        """
+        self._add_components(components, init)
+        return self
+
+    def _add_components(self, components, init):
+        if isinstance(components, str):
+            components = (components,)
+            init = (init,)
+        elif isinstance(components, (tuple, list)):
+            components = tuple(components)
+            if init is None:
+                init = (None,) * len(components)
+            else:
+                init = tuple(init)
+
+        for comp, value in zip(components, init):
+            if hasattr(self, comp):
+                raise ValueError("An attribute '%s' already exists" % comp)
+            if self.components is not None and comp in self.components:
+                raise ValueError("A components '%s' already exists" % comp)
+
+            if self.components is None:
+                self.components = tuple([comp])
+                if self._data is not None:
+                    warnings.warn("All batch data is erased")
+            else:
+                self.components = self.components + tuple([comp])
+            setattr(self, comp, value)
+
+            self.meta[comp] = dict()
+
+    def update_component(self, component, value):
+        """ Add new component or update existed one
+
+        Parameters
+        ----------
+        component : str
+            component name
+        value : array-like
+            component data
+        """
+        if component not in self.components:
+            self._add_components(component, init=value)
+        else:
+            setattr(self, component, value)
 
     #-------------------------------------------------------------------------#
     #                              Load and Dump                              #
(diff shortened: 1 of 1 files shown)


seismicpro/src/seismic_batch.py
```diff
@@ -257,18 +257,80 @@ def copy_meta(self, from_comp, to_comp):
...
-                raise ValueError('{} does not exist.'.format(fr_comp))
+                raise ValueError('Meta of the component {} does not exist.'.format(fr_comp))
 
             if fr_comp == t_comp:
                 continue
 
-            if self.meta[t_comp]:
-                warnings.warn("Meta of component {} is not empty and".format(t_comp) + \
-                              " will be replaced by the meta from component {}.".format(fr_comp),
+            if self.meta.get(t_comp):
```
nikita-klsh, line 265: I believe the condition `if t_comp in self.meta` is equivalent but more explicit

seismicpro/src/seismic_batch.py
```diff
@@ -257,18 +257,80 @@ def copy_meta(self, from_comp, to_comp):
...
+                warnings.warn("Meta of the component {} is not empty and".format(t_comp) + \
+                              " will be replaced by the meta from the component {}.".format(fr_comp),
                               UserWarning)
+
             self.meta[t_comp] = self.meta[fr_comp].copy()
         return self
 
+    @action
+    def add_components(self, components, init=None):
```
nikita-klsh, line 274: I would suggest reusing the existing code.   
Let the `batch.add_components` initialize the component. And `seismic_batch.add_components` only initialize the meta.

``` 
super().add_components(components, init) 
self.meta[components] = {}
```

seismicpro/src/seismic_batch.py
```diff
@@ -257,18 +257,80 @@ def copy_meta(self, from_comp, to_comp):
...
+                self.components = tuple([comp])
+                if self._data is not None:
+                    warnings.warn("All batch data is erased")
+            else:
+                self.components = self.components + tuple([comp])
+            setattr(self, comp, value)
+            if comp not in self.meta:
+                self.meta[comp] = dict()
+
+    def update_component(self, component, value):
```
nikita-klsh, line 319: Why do we need this when there is `pipeline.update()`?

seismicpro/src/seismic_batch.py
```diff
@@ -1178,7 +1242,7 @@ def mcm(self, src, dst, eps=3, length_win=12):
         long_win, lead_win = energy, energy
         lead_win[:, length_win:] = lead_win[:, length_win:] - lead_win[:, :-length_win]
         energy = lead_win / (long_win + eps)
-        self.add_components(dst, init=np.array(energy + [None])[:-1]) # array implicitly converted to object dtype
+        self.update_component(dst, np.array(energy + [None])[:-1]) # array implicitly converted to object dtype
```
nikita-klsh, line 1245: Why do you use `update` instead of `add` here?
It looks more like the component is created here, not updated

seismicpro/src/seismic_batch.py
anton-br: no, because meta for t_comp can be empty and my condition will catch this case while your not.

seismicpro/src/seismic_batch.py
anton-br: because sometimes we want to create component in pipeline. In this case, `add_component` will return an error. Update component will only change the data inside the component

seismicpro/src/seismic_batch.py
```diff
@@ -222,6 +222,45 @@ def _post_filter_by_mask(self, mask, *args, **kwargs):
...
+        super().add_components(components, init)
+
+        components = (components,) if isinstance(components, str) else components
+        for comp in components:
+            if comp not in self.meta:
+                self.meta[comp] = dict()
+        return self
+
+    def update_component(self, component, value):
+        """ Add new component or update existed one
```
a-arefina, line 250:
```suggestion
        """ Add a new component or update an existing one
```

seismicpro/src/seismic_batch.py
```diff
@@ -257,15 +296,16 @@ def copy_meta(self, from_comp, to_comp):
...
-                raise ValueError('{} does not exist.'.format(fr_comp))
+                raise ValueError('Meta of the component {} does not exist.'.format(fr_comp))
 
             if fr_comp == t_comp:
                 continue
 
-            if self.meta[t_comp]:
-                warnings.warn("Meta of component {} is not empty and".format(t_comp) + \
-                              " will be replaced by the meta from component {}.".format(fr_comp),
+            if t_comp in self.meta and self.meta[t_comp] != dict():
```
a-arefina, line 304:
```suggestion
            if self.meta.get(t_comp, None):
```

empty dict evaluates to false

anton-br: please, decide with @nikita-klsh which approach is better..

a-arefina (approved):

dpodvyaznikov (approved):

nikita-klsh (approved):

seismicpro/src/seismic_batch.py
```diff
@@ -222,6 +222,45 @@ def _post_filter_by_mask(self, mask, *args, **kwargs):
...
+
+        Raises
+        ------
+        ValueError
+            If a component or an attribute with the given name already exists
+        """
+        super().add_components(components, init)
+
+        components = (components,) if isinstance(components, str) else components
+        for comp in components:
```
alexandrkuvaev, line 244:
```python
for comp in np.array(components)
```
instead of these two lines?

seismicpro/src/seismic_batch.py
```diff
@@ -222,6 +222,45 @@ def _post_filter_by_mask(self, mask, *args, **kwargs):
...
+            if comp not in self.meta:
+                self.meta[comp] = dict()
+        return self
+
+    def update_component(self, component, value):
```
alexandrkuvaev, line 249: Why `add_components` returns `self`, while this method doesn't?

alexandrkuvaev (changes requested):

PR #50, opened 2021-03-17
Requirements

### LINKED ISSUE #43: Update install and requirements
Check dependencies and their versions

a-arefina:
fixes #43 

* PyWavelets removed from code and requirements
* setup.py updated to match requirements and also includes batchflow (there might be a better way to indicate the bf revision)
* test data moved inside package to allow running tests from an installed package


a-arefina:
not quite sure if both install_with_pip and install_with_pipenv are really needed

requirements.txt
```diff
@@ -1 +1,9 @@
-.
\ No newline at end of file
+matplotlib>=3.3.1
```
roman-kh, line 1: AFAIR, you can put `.` in req.txt, in order not to duplicate `install_requires`

setup.py
```diff
@@ -18,19 +18,23 @@
...
+        'scikit-learn>=0.23.2',
+        'scipy>=1.5.2',
+        'segyio>=1.9.5',
         'tdigest>=0.5.2.2',
-        'tqdm==4.30.0',
-        'segyio==1.8.3',
-        'scikit-image>=0.13.1',
-        'numba>=0.35.0'
+        'tqdm>=4.56.0',
+        'batchflow @ git+https://github.com/analysiscenter/batchflow.git@9823f369#egg=batchflow',
```
roman-kh, line 33: It seems pretty legit to me

setup.py
```diff
@@ -18,19 +18,23 @@
...
+
+        # 'dill>=0.2.7.1',
+        # 'scikit-image>=0.13.1',
+        # 'numba>=0.35.0'
     ],
     extras_require={
         'tensorflow': ['tensorflow>=1.12'],
```
roman-kh, line 40: Do we need `tf`? `2.2` > `1.12`

roman-kh:
> not quite sure if both install_with_pip and install_with_pipenv are really needed

Install order might differ, so it's better to reveal problems in advance.

.github/workflows/test-install.yml
```diff
@@ -0,0 +1,126 @@
...
+
+# -----------------------------------------
+#    Use a module from local source
+# -----------------------------------------
+  use_as_local_module:
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest] #, macos-latest, windows-latest]
```
nikita-klsh, line 18: Should we consider the loop over all OS's since we have classifier `Operating System :: OS Independent`?

roman-kh: Even though pytorch supports ROCm, the latter does not support MacOS yet. So we can definitely exclude MacOS for now.

dpodvyaznikov (approved):

dpodvyaznikov:
If there are no objections, I'd merge this by the end of the day.

PR #56, opened 2021-07-27
reindex improvements

alexandrkuvaev:
* Fix `SeismicIndex.reindex`:
    * Now iteration is allowed over reindexed instance,
    * Now it's possible to optionally reindex all nested indices (`train`, `test` and `validation`),
    * Reindexing of underlying surveys is now optional.
* Add reindex to `SeismicDataset` which simply reindexes its `index` attribute

### DIFF
--- a/seismicpro/src/dataset.py
+++ b/seismicpro/src/dataset.py
@@ -109,7 +109,7 @@ class SeismicDataset(Dataset):
         return type(self).from_dataset(self, index)
 
     def reindex(self, new_index, reindex_nested=False, reindex_surveys=False):
-        """Reindex the index of `self` with new headers columns.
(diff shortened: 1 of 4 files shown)


seismicpro/src/dataset.py
```diff
@@ -108,6 +108,27 @@ def create_subset(self, index):
             index = self.index.create_subset(index)
         return type(self).from_dataset(self, index)
 
+    def reindex(self, new_index, reindex_nested=False, reindex_surveys=False):
+        """Reindex the index of `self` with new headers columns.
```
bestetc, line 112: It is better to write about inplace only execution in docstring.
One word in the return section couldn't be enough to explicit this.

anton-br (approved):

bestetc (approved):

PR #60, opened 2021-08-16
QC for stacking velocities

nikita-klsh:
Adds functionality to QC velocube, either the stored velocities(read from the file) or interpolated velocities.


alexandrkuvaev:
Covered in https://github.com/gazprom-neft/SeismicPro/pull/63, closed

PR #63, opened 2021-09-22
Implement stacking velocities QC

alexandrkuvaev:
Implements:
* 3 types of stacking velocities QC
* njitted 1d interpolator: up to 3 times faster than `scipy` version
* Stacking velocity interpolation for multiple spatial coordinates over a regular grid of times

### DIFF
--- a/seismicpro/src/utils/interpolation.py
+++ b/seismicpro/src/utils/interpolation.py
@@ -17,6 +17,7 @@ def interpolate(x_new, x, y, left_slope, right_slope):
     return res
 
 
+#pylint: disable=invalid-name
 class interp1d:
     """Return a 1d piecewise linear interpolant to a function defined by pairs of data points `(x, y)`. Function values
     at points outside the `x` range will be linearly extrapolated.
--- a/seismicpro/src/utils/interpolation.py
+++ b/seismicpro/src/utils/interpolation.py
@@ -19,16 +19,17 @@ class interp1d:
         y = np.array(y)
 
         if len(x) < 2:
-            raise ValueError
+            raise ValueError("At least two points should be passed to perform interpolation")
 
         ind = np.argsort(x, kind="mergesort")
         self.x = x[ind]
         self.y = y[ind]
 
-        self.left_slope = (y[1] - y[0]) / (x[1] - x[0])
-        self.right_slope = (y[-1] - y[-2]) / (x[-1] - x[-2])
+        self.left_slope = (self.y[1] - self.y[0]) / (self.x[1] - self.x[0])
+        self.right_slope = (self.y[-1] - self.y[-2]) / (self.x[-1] - self.x[-2])
 
     def __call__(self, x):
         x = np.array(x)
-        res = interpolate(x, self.x, self.y, self.left_slope, self.right_slope)
-        return res.item() if x.ndim == 0 else res
+        is_scalar_input = (x.ndim == 0)
+        res = interpolate(x.ravel(), self.x, self.y, self.left_slope, self.right_slope)
+        return res.item() if is_scalar_input else res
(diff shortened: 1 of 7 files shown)


seismicpro/src/velocity_cube.py
```diff
@@ -477,6 +533,9 @@ def create_interpolator(self):
         if not self.stacking_velocities_dict:
             raise ValueError("No stacking velocities passed")
         self.interpolator = VelocityInterpolator(self.stacking_velocities_dict)
+        # Precompile numba functions inside interpolator
+        coords_center = self.interpolator.coords.mean(axis=0, keepdims=True)
+        _ = self.interpolator.interpolate(coords=coords_center, times=np.zeros(1))
```
nikita-klsh, line 538: Do you compile the right version? Here times.dtype is float64.
I guess later it will be called with gather.times, dtype float32 or np,arange(0, 6000, 2), dtype int64

seismicpro/src/velocity_cube.py
```diff
@@ -510,3 +569,38 @@ def __call__(self, inline, crossline, create_interpolator=True):
...
         return self.interpolator(inline, crossline)
+
+    def qc(self, win_radius, times, coords=None, metrics_names=None, n_workers=None):
+        if metrics_names is None:
+            metrics_names = velocity_qc.VELOCITY_QC_METRICS
+        metrics_funcs = []
+        for name in to_list(metrics_names):
+            if name in velocity_qc.VELOCITY_QC_METRICS:
+                func = getattr(velocity_qc, name)
+            elif callable(name):
```
nikita-klsh, line 580: Currently this case is not working correctly

alexandrkuvaev: Fixed

seismicpro/src/velocity_cube.py
alexandrkuvaev: Removed precompilation completely. Probably it will affect the seismic cube stacking pipeline without preliminary warmup. If so, we will think of it later since it's not trivial to fix for the general case.

seismicpro/src/velocity_cube.py
```diff
@@ -510,3 +593,80 @@ def __call__(self, inline, crossline, create_interpolator=True):
             if self.is_dirty_interpolator:
                 warnings.warn("Dirty interpolator is being used", RuntimeWarning)
         return self.interpolator(inline, crossline)
+
+    def qc(self, win_radius, times, coords=None, metrics_names=None, n_workers=None):  #pylint: disable=invalid-name
+        """Perform quality control of the velocity cube by calculating spatial-window-based metrics for stacking
+        velocities evaluated at given `times`.
+
+        The QC is performed by calculating spatial-window-based metrics for stacking velocities defined in
+        `metrics_names`. + times, coords, returns metrics map
```
nikita-klsh, line 602: Strange paragraph. The first sentence is a heavy repetition of the previous one. The second one looks more like a code

seismicpro/src/utils/interpolation.py
```diff
@@ -0,0 +1,62 @@
+"""Implements a class for linear 1d interpolation and extrapolation"""
+
+import numpy as np
+from numba import njit
+
+
+@njit(nogil=True)
+def interpolate(x_new, x, y, left_slope, right_slope):
+    """Return a 1d piecewise linear interpolant to a function defined by pairs of data points `(x, y)`, evaluated at
```
nikita-klsh, line 9: Here you return the result of interpolation, not interpolant

alexandrkuvaev: Return interpolant evaluated at ...

seismicpro/src/velocity_cube.py
alexandrkuvaev: Removed

dpodvyaznikov (approved):

PR #69, opened 2021-11-19
Getitem

anton-br:
This pr provide a slicing, advanced indexing and copy functional.

### DIFF
--- a/seismicpro/src/gather.py
+++ b/seismicpro/src/gather.py
@@ -2,7 +2,7 @@
 
 import os
 import warnings
-from copy import deepcopy
+from copy import copy, deepcopy
 from textwrap import dedent
 
 import segyio
@@ -91,19 +91,40 @@ class Gather:
         return self.data.shape
 
     def __getitem__(self, key):
-        """Select gather headers by their names.
+            """!!Select gather headers by their names.
 
-        Parameters
-        ----------
-        key : str or list of str
-            Gather headers to get.
+            Parameters
+            ----------
+            key : str, list of str, list of slices/int
+                IF ..... Gather headers to get.
 
-        Returns
-        -------
-        headers : np.ndarray
-            Headers values.
-        """
-        return self.headers[key].values
+            Returns
+            -------
+            headers : np.ndarray
+                Headers values.
+            """
+            # If key is string or array of str, return header columns with specified names
+            keys_array = np.array(to_list(key))
+            if keys_array.dtype.type == np.str_:
+                self.validate(required_header_cols=keys_array)
+                return self.headers[keys_array].values
+
+            # Other case is to slice gather with given key. Here key can be any type that might be processed by
+            # np.array's getitem.
+            key = to_list(key)
+            key = key + [slice(None)] if len(key) == 1 else key
+            key = tuple([slice(k, k+1) if isinstance(k, (int, np.integer)) else k for k in key])
+
+            new_self = copy(self)
+            new_self.data = self.data[key]
+            if new_self.data.size == 0:
+                raise ValueError('!!')
+
+            # The first axis is responsible for the number of traces, so we have to process traces headers.
+            # The second axis is responsible for time, we need to process traces time descriptions.
+            new_self.headers = self.headers.iloc[key[0]]
+            new_self.samples = self.samples[key[1]]
+            return new_self
 
     def __setitem__(self, key, value):
         """Set given values to selected gather headers.
@@ -205,6 +226,12 @@ class Gather:
         self_copy.survey = survey
         self.survey = survey
         return self_copy
+
+    @batch_method
+    def getitem(self, *args, copy=False):
+        """!!"""
+        new_self = self.copy() if copy else self
+        return new_self[args]
 
     def _validate_header_cols(self, required_header_cols):
         """Check if the gather headers contain all columns from `required_header_cols`."""
--- a/seismicpro/src/gather.py
+++ b/seismicpro/src/gather.py
@@ -115,7 +115,7 @@ class Gather:
             key = key + [slice(None)] if len(key) == 1 else key
             key = tuple([slice(k, k+1) if isinstance(k, (int, np.integer)) else k for k in key])
 
-            new_self = copy(self)
+            new_self = self.copy(except_attrs=['data', 'headers'])
             new_self.data = self.data[key]
             if new_self.data.size == 0:
                 raise ValueError('!!')
@@ -212,19 +212,36 @@ class Gather:
         return tuple(coords[0].tolist())
 
     @batch_method(target='threads', copy_src=False)
-    def copy(self):
-        """Perform a deepcopy of all gather attributes except for `survey`, which is kept unchanged.
+    def copy(self, except_attrs=None):
+        """Perform a deepcopy of all gather attributes except for `survey` and `except_attrs`, which are kept
+        unchanged.
+
+        Parameters
+        ----------
+        except_attrs : str or array of str, default None
+            Attributes that will be kept unchanged after the copy.
 
         Returns
         -------
         copy : Gather
             Copy of the gather.
         """
-        survey = self.survey
-        self.survey = None
+        # To avoid copying `except_attrs` and `survey`, save thier values in the dict and None them in
+        # the souce object.
+        except_attrs = [] if except_attrs is None else to_list(except_attrs)
+        except_list = except_attrs + ['survey']
+
+        nocopy_dict = {}
+        for attr in except_list:
+            nocopy_dict[attr] = getattr(self, attr)
+            setattr(self, attr, None)
+
+        # Make a deepcopy of self and set back unchanged attributes
         self_copy = deepcopy(self)
-        self_copy.survey = survey
-        self.survey = survey
+        self_copy.__dict__.update(nocopy_dict)
+
+        # Reset all None attrs in self
+        {setattr(self, attr, value) for attr, value in nocopy_dict.items()}
         return self_copy
 
     @batch_method
--- a/seismicpro/src/gather.py
+++ b/seismicpro/src/gather.py
@@ -2,7 +2,7 @@
 
 import os
 import warnings
-from copy import copy, deepcopy
+from copy import deepcopy
 from textwrap import dedent
 
 import segyio
@@ -91,40 +91,40 @@ class Gather:
         return self.data.shape
 
     def __getitem__(self, key):
-            """!!Select gather headers by their names.
+        """!!Select gather headers by their names.
 
-            Parameters
-            ----------
-            key : str, list of str, list of slices/int
-                IF ..... Gather headers to get.
+        Parameters
+        ----------
+        key : str, list of str, list of slices/int
+            IF ..... Gather headers to get.
 
-            Returns
-            -------
-            headers : np.ndarray
-                Headers values.
-            """
-            # If key is string or array of str, return header columns with specified names
-            keys_array = np.array(to_list(key))
-            if keys_array.dtype.type == np.str_:
-                self.validate(required_header_cols=keys_array)
-                return self.headers[keys_array].values
+        Returns
+        -------
+        headers : np.ndarray
+            Headers values.
+        """
+        # If key is string or array of str, return header columns with specified names
+        keys_array = np.array(to_list(key))
+        if keys_array.dtype.type == np.str_:
+            self.validate(required_header_cols=keys_array)
+            return self.headers[keys_array].values
 
-            # Other case is to slice gather with given key. Here key can be any type that might be processed by
-            # np.array's getitem.
-            key = to_list(key)
-            key = key + [slice(None)] if len(key) == 1 else key
-            key = tuple([slice(k, k+1) if isinstance(k, (int, np.integer)) else k for k in key])
+        # Other case is to slice gather with given key. Here key can be any type that might be processed by
+        # np.array's getitem.
+        key = to_list(key)
+        key = key + [slice(None)] if len(key) == 1 else key
+        key = tuple(slice(k, k+1) if isinstance(k, (int, np.integer)) else k for k in key)
 
-            new_self = self.copy(except_attrs=['data', 'headers'])
-            new_self.data = self.data[key]
-            if new_self.data.size == 0:
-                raise ValueError('!!')
+        new_self = self.copy(except_attrs=['data', 'headers'])
+        new_self.data = self.data[key]
+        if new_self.data.size == 0:
+            raise ValueError('!!')
 
-            # The first axis is responsible for the number of traces, so we have to process traces headers.
-            # The second axis is responsible for time, we need to process traces time descriptions.
-            new_self.headers = self.headers.iloc[key[0]]
-            new_self.samples = self.samples[key[1]]
-            return new_self
+        # The first axis is responsible for the number of traces, so we have to process traces headers.
+        # The second axis is responsible for time, we need to process traces time descriptions.
+        new_self.headers = self.headers.iloc[key[0]]
+        new_self.samples = self.samples[key[1]]
+        return new_self
 
     def __setitem__(self, key, value):
         """Set given values to selected gather headers.
@@ -241,7 +241,8 @@ class Gather:
         self_copy.__dict__.update(nocopy_dict)
 
         # Reset all None attrs in self
-        {setattr(self, attr, value) for attr, value in nocopy_dict.items()}
+        for attr, value in nocopy_dict.items():
+            setattr(self, attr, value)
         return self_copy
 
     @batch_method


seismicpro/src/gather.py
```diff
@@ -91,19 +91,40 @@ def shape(self):
...
+        new_self = self.copy(except_attrs=['data', 'headers'])
+        new_self.data = self.data[key]
+        if new_self.data.size == 0:
+            raise ValueError('!!')
+
+        # The first axis is responsible for the number of traces, so we have to process traces headers.
+        # The second axis is responsible for time, we need to process traces time descriptions.
+        new_self.headers = self.headers.iloc[key[0]]
+        new_self.samples = self.samples[key[1]]
+        return new_self
```
alexandrkuvaev, line 127: Don't forget to update `sort_by`

seismicpro/src/gather.py
```diff
@@ -91,19 +91,40 @@ def shape(self):
...
-        return self.headers[key].values
+        # If key is string or array of str, return header columns with specified names
+        keys_array = np.array(to_list(key))
+        if keys_array.dtype.type == np.str_:
+            self.validate(required_header_cols=keys_array)
+            return self.headers[keys_array].values
+
+        # Other case is to slice gather with given key. Here key can be any type that might be processed by
+        # np.array's getitem.
+        key = to_list(key)
```
alexandrkuvaev, line 114: Let's call it just once in the beginning and then check for key elements types

seismicpro/src/gather.py
```diff
@@ -191,21 +212,45 @@ def get_coords(self, coords_columns="index"):
         return tuple(coords[0].tolist())
 
     @batch_method(target='threads', copy_src=False)
-    def copy(self):
-        """Perform a deepcopy of all gather attributes except for `survey`, which is kept unchanged.
+    def copy(self, except_attrs=None):
```
alexandrkuvaev, line 215: Simply `except`

seismicpro/src/gather.py
```diff
@@ -191,21 +212,45 @@ def get_coords(self, coords_columns="index"):
...
         -------
         copy : Gather
             Copy of the gather.
         """
-        survey = self.survey
-        self.survey = None
+        # To avoid copying `except_attrs` and `survey`, save thier values in the dict and None them in
+        # the souce object.
+        except_attrs = [] if except_attrs is None else to_list(except_attrs)
+        except_list = except_attrs + ['survey']
```
alexandrkuvaev, line 232: Possible double `survey`, better make it a set

seismicpro/src/gather.py
```diff
@@ -191,21 +212,45 @@ def get_coords(self, coords_columns="index"):
...
+
+        # Make a deepcopy of self and set back unchanged attributes
         self_copy = deepcopy(self)
-        self_copy.survey = survey
-        self.survey = survey
+        self_copy.__dict__.update(nocopy_dict)
+
+        # Reset all None attrs in self
+        for attr, value in nocopy_dict.items():
+            setattr(self, attr, value)
```
alexandrkuvaev, line 245: Why not `self.__dict__.update(nocopy_dict)`

seismicpro/src/gather.py
```diff
@@ -191,21 +212,45 @@ def get_coords(self, coords_columns="index"):
...
         return self_copy
 
+    @batch_method
+    def getitem(self, *args, copy=False):
+        """!!"""
+        new_self = self.copy() if copy else self
+        return new_self[args]
```
alexandrkuvaev, line 252: I'd copy the result of `__getitem__` since it'll almost always be more efficient

seismicpro/src/gather.py
```diff
@@ -191,21 +212,45 @@ def get_coords(self, coords_columns="index"):
...
-        self.survey = None
+        # To avoid copying `except_attrs` and `survey`, save thier values in the dict and None them in
+        # the souce object.
+        except_attrs = [] if except_attrs is None else to_list(except_attrs)
+        except_list = except_attrs + ['survey']
+
+        nocopy_dict = {}
+        for attr in except_list:
+            nocopy_dict[attr] = getattr(self, attr)
+            setattr(self, attr, None)
```
alexandrkuvaev, line 237: Try avoiding this by creating a `memo` dict here to be passed to `deepcopy`

alexandrkuvaev (approved):

seismicpro/src/gather.py
```diff
@@ -91,19 +91,53 @@ def shape(self):
         return self.data.shape
 
     def __getitem__(self, key):
-        """Select gather headers by their names.
+        """Depend on type of `key` either select gather headers by their names or create a new instance of Gather with
+        specified subset of traces.
```
alexandrkuvaev, line 95: I'd say, traces and samples

seismicpro/src/gather.py
```diff
@@ -91,19 +91,53 @@ def shape(self):
         return self.data.shape
 
     def __getitem__(self, key):
-        """Select gather headers by their names.
+        """Depend on type of `key` either select gather headers by their names or create a new instance of Gather with
```
alexandrkuvaev, line 94: Depending

seismicpro/src/gather.py
```diff
@@ -91,19 +91,53 @@ def shape(self):
...
+        -----
+        If the data after getitem is no longer sorted, `sort_by` attribute in resulted Gather will be set to None.
 
         Parameters
         ----------
-        key : str or list of str
-            Gather headers to get.
+        key : str, list of str, int, list, tuple, slice
+            If str or list of str, gather headers to get.
+            Otherwise, indices of traces to get. Here, getitem behaviour coinside with getitem in numpy.array.
```
alexandrkuvaev, line 105: Coincides, also mention samples selection

seismicpro/src/gather.py
```diff
@@ -91,19 +91,53 @@ def shape(self):
...
-            Headers values.
+        result : np.ndarray or Gather
+            Headers values or Gather with specified subset of traces.
         """
-        return self.headers[key].values
+        # If key is string or array of str, return header columns with specified names
+        keys_array = np.array(to_list(key))
+        if keys_array.dtype.type == np.str_:
+            self.validate(required_header_cols=keys_array)
+            # We avoid to use pandas indexing with multiple columns due to the performance reasons.
```
alexandrkuvaev, line 116: Avoid using, to speed up selection of headers from gathers with small number of traces

seismicpro/src/gather.py
```diff
@@ -91,19 +91,53 @@ def shape(self):
...
+            return np.column_stack([self.headers[col].values if col in self.headers.columns
```
alexandrkuvaev, line 117: Consider avoiding one-lining, looks scary here

seismicpro/src/gather.py
```diff
@@ -91,19 +91,53 @@ def shape(self):
...
+                                    else self.headers.index.get_level_values(col).values for col in keys_array])
+
+        # Other case is to slice gather with given key. Here key can be any type that might be processed by
+        # np.array's getitem.
+        key = [key] if not isinstance(key, tuple) else key
```
alexandrkuvaev, line 122: Let's make it `tuple` then here

seismicpro/src/gather.py
```diff
@@ -91,19 +91,53 @@ def shape(self):
...
+        key = list(key) + [slice(None)] if len(key) == 1 else key
+        key = tuple(slice(k, k+1) if isinstance(k, (int, np.integer)) else k for k in key)
+
+        new_self = self.copy(ignore=['data', 'headers'])
```
alexandrkuvaev, line 126: And `samples`

seismicpro/src/gather.py
```diff
@@ -91,19 +91,53 @@ def shape(self):
...
+
+        new_self.data = np.atleast_2d(self.data[key])
```
alexandrkuvaev, line 128: It'll result in wrong data if a single sample for several traces was chosen

seismicpro/src/gather.py
```diff
@@ -191,26 +225,39 @@ def get_coords(self, coords_columns="index"):
         return tuple(coords[0].tolist())
 
     @batch_method(target='threads', copy_src=False)
-    def copy(self):
-        """Perform a deepcopy of all gather attributes except for `survey`, which is kept unchanged.
+    def copy(self, ignore=None):
+        """Perform a deepcopy of all gather attributes except for `survey` and `ignore`, which are kept
```
alexandrkuvaev, line 229: And those, specified in `ignore`

seismicpro/src/gather.py
```diff
@@ -191,26 +225,39 @@ def get_coords(self, coords_columns="index"):
...
-        survey = self.survey
-        self.survey = None
-        self_copy = deepcopy(self)
-        self_copy.survey = survey
-        self.survey = survey
-        return self_copy
+        # To avoid copying `ignore` and `survey`, save thier values in the dict and None them in
+        # the souce object.
+        ignore_attrs = set() if ignore is None else set(to_list(ignore))
+        ignore_attrs = [getattr(self, attr) for attr in ignore_attrs | set(['survey'])]
```
alexandrkuvaev, line 245: Why not `{"survey"}`?

seismicpro/src/gather.py
```diff
@@ -191,26 +225,39 @@ def get_coords(self, coords_columns="index"):
...
         copy : Gather
             Copy of the gather.
         """
-        survey = self.survey
-        self.survey = None
-        self_copy = deepcopy(self)
-        self_copy.survey = survey
-        self.survey = survey
-        return self_copy
+        # To avoid copying `ignore` and `survey`, save thier values in the dict and None them in
```
alexandrkuvaev, line 242: ?

seismicpro/src/gather.py
```diff
@@ -191,26 +225,39 @@ def get_coords(self, coords_columns="index"):
...
+        """A public interface for `self.__getitem__` method."""
+        return self[args]
 
     def _validate_header_cols(self, required_header_cols):
         """Check if the gather headers contain all columns from `required_header_cols`."""
-        headers = self.headers.reset_index()
+        header_cols = set(self.headers.columns) | set(self.headers.index.names)
         required_header_cols = to_list(required_header_cols)
-        if any(col not in headers for col in required_header_cols):
+        if any(col not in header_cols for col in required_header_cols):
```
alexandrkuvaev, line 260: Now you can simply calc set difference

seismicpro/src/gather.py
```diff
@@ -191,26 +225,39 @@ def get_coords(self, coords_columns="index"):
...
+        ignore_attrs = set() if ignore is None else set(to_list(ignore))
+        ignore_attrs = [getattr(self, attr) for attr in ignore_attrs | set(['survey'])]
+
+        # Construct a memo dict with attributes to avoid thier copying during deepcopy.
+        memo = {id(attr): attr for attr in ignore_attrs}
+        return deepcopy(self, memo)
+
+    @batch_method()
+    def getitem(self, *args):
+        """A public interface for `self.__getitem__` method."""
```
alexandrkuvaev, line 253: It's not "public"

alexandrkuvaev (changes requested):

seismicpro/src/gather.py
anton-br: what about now?

seismicpro/src/gather.py
anton-br: done, have a look

seismicpro/src/gather.py
```diff
@@ -91,19 +91,63 @@ def shape(self):
         return self.data.shape
 
     def __getitem__(self, key):
-        """Select gather headers by their names.
+        """Depending on the `key` type, either select gather headers by their names or create a new instance of Gather
+        with specified traces and samples.
+
+        Notes
+        -----
+        If the data after getitem is no longer sorted, `sort_by` attribute in resulted Gather will be set to None.
```
alexandrkuvaev, line 99:
```suggestion
        If the data after `__getitem__` is no longer sorted, `sort_by` attribute in the resulting `Gather` will be set to `None`.
```

seismicpro/src/gather.py
```diff
@@ -91,19 +91,63 @@ def shape(self):
         return self.data.shape
 
     def __getitem__(self, key):
-        """Select gather headers by their names.
+        """Depending on the `key` type, either select gather headers by their names or create a new instance of Gather
+        with specified traces and samples.
+
+        Notes
```
alexandrkuvaev, line 97: Add a note, that if header selection is performed, a 2d array is always returned even for a single header

seismicpro/src/gather.py
```diff
@@ -91,19 +91,63 @@ def shape(self):
         return self.data.shape
 
     def __getitem__(self, key):
-        """Select gather headers by their names.
+        """Depending on the `key` type, either select gather headers by their names or create a new instance of Gather
+        with specified traces and samples.
```
alexandrkuvaev, line 95: Either select gather headers by their names or create a new `Gather` with specified traces and samples depending on the `key` type.

seismicpro/src/gather.py
```diff
@@ -91,19 +91,63 @@ def shape(self):
...
         Returns
         -------
-        headers : np.ndarray
-            Headers values.
+        result : np.ndarray or Gather
+            Headers values or Gather with a specified subset of traces.
         """
-        return self.headers[key].values
+        # If key is string or array of str, return header columns with specified names
+        keys_array = np.array(to_list(key))
```
alexandrkuvaev, line 114: Here you can probably perform just `np.array(key)`

seismicpro/src/gather.py
```diff
@@ -91,19 +91,63 @@ def shape(self):
...
+            headers = []
+            for col in keys_array:
+                header = self.headers[col] if col in self.headers.columns else self.headers.index.get_level_values(col)
+                headers.append(header.values)
+            return np.column_stack(headers)
+
+        # Another case is to slice gather with the given key. Here key can be any type that might be processed by
+        # np.array's getitem.
+        key = (key, ) if not isinstance(key, tuple) else key
+        key =  key + (slice(None), ) if len(key) == 1 else key
```
alexandrkuvaev, line 128:
```suggestion
        key = key + (slice(None), ) if len(key) == 1 else key
```

seismicpro/src/gather.py
```diff
@@ -91,19 +91,63 @@ def shape(self):
...
+        new_self = self.copy(ignore=['data', 'headers', 'samples'])
+        new_self.data = self.data[key_tuple]
+        if new_self.data.ndim != 2:
+            raise ValueError("Given `key` leads either to ambiguous behavior when processing samples and headers or to"
+                             " a change in the dimension of the data")
+        if new_self.data.size == 0:
+            raise ValueError("Given `key` results in empty object")
+
+        # The first axis is responsible for the number of traces, so we have to process traces headers.
+        # The second axis is responsible for time. We need to process traces time descriptions.
```
alexandrkuvaev, line 143: Simplify comment

seismicpro/src/gather.py
```diff
@@ -91,19 +91,63 @@ def shape(self):
...
+        new_self.headers = self.headers.iloc[key_tuple[0]]
+        new_self.samples = self.samples[key_tuple[1]]
+
+        # Check that `sort_by` is still represents the actual trace sorting since it may be changed during getitem.
```
alexandrkuvaev, line 147: Grammarly

seismicpro/src/gather.py
```diff
@@ -191,28 +235,39 @@ def get_coords(self, coords_columns="index"):
...
-    def copy(self):
-        """Perform a deepcopy of all gather attributes except for `survey`, which is kept unchanged.
+    def copy(self, ignore=None):
+        """Perform a deepcopy of all gather attributes except for `survey` and those specified in ignore, which are
+        kept unchanged.
+
+        Parameters
+        ----------
+        ignore : str or array of str, default None
+            Attributes that will be kept unchanged after the copy.
```
alexandrkuvaev, line 245:
```suggestion
            Attributes that won't be copied.
```

alexandrkuvaev (changes requested):

seismicpro/src/gather.py
anton-br: you can't, since `np.array('offset')` result in 0-dim array. And you cannot iterate over 0-dim array(

seismicpro/src/gather.py
anton-br: done

seismicpro/src/gather.py
anton-br: done

seismicpro/src/gather.py
anton-br: done

alexandrkuvaev (approved):

seismicpro/src/gather.py
```diff
@@ -191,28 +241,39 @@ def get_coords(self, coords_columns="index"):
...
-        self.survey = survey
-        return self_copy
+        ignore_attrs = set() if ignore is None else set(to_list(ignore))
+        ignore_attrs = [getattr(self, attr) for attr in ignore_attrs | {'survey'}]
+
+        # Construct a memo dict with attributes to avoid their copying during deepcopy.
+        memo = {id(attr): attr for attr in ignore_attrs}
+        return deepcopy(self, memo)
+
+    @batch_method()
```
alexandrkuvaev, line 265: Specify `target` explicitly, please

alexandrkuvaev (changes requested):

seismicpro/src/tests/gather_test.py
```diff
@@ -1,43 +1,160 @@
...
+
+key_types = [0, (0, ), (0, 1), [0], [0, 4], slice(None, 5, None)]
+fail_keys = list(product([(0, 1), (0, ), [0], [0, 4]], repeat=2))
+@pytest.mark.parametrize('key', key_types + list(product(key_types, repeat=2)))
+def test_gather_getitem_gathers(gather, key):
+    """test_gather_getitem_gathers"""
+    if key in fail_keys:
+        pytest.raises(ValueError, gather.__getitem__, key)
+        pytest.raises(ValueError, gather.get_item, key)
+        pytest.xfail()
```
a-arefina, line 86: What about extracting this to a separate `test_gather_getitem_gathers_fail` ?

seismicpro/src/tests/gather_test.py
```diff
@@ -1,43 +1,160 @@
...
+    result_get_item = gather.get_item(key)
+    expected_data = gather.data[key]
+
+    compare_gathers(result_getitem, result_get_item, check_types=True)
+    assert np.allclose(result_getitem.data.reshape(-1), expected_data.reshape(-1))
+
+    # Find a correct shape of data when numpy indexing works differently
+    keys = (key, ) if not isinstance(key, tuple) else key
+    if result_getitem.shape != expected_data.shape:
+        expected_shape = tuple(1 if isinstance(k, int) else k.stop if isinstance(k, slice) else len(k) for k in keys)
```
a-arefina, line 98: please add braces to make operation priorities more clear. Imo, this expression is a bit hard to be parced by a human :(

seismicpro/src/tests/gather_test.py
```diff
@@ -1,43 +1,160 @@
...
+        if len(expected_shape) < 2:
+            expected_shape = expected_shape + (gather.shape[1], )
+        assert result_getitem.shape == expected_shape, f"for the key {key} expected {result_getitem.shape} shape but "\
+                                                       f"received {expected_shape}"
+
+    assert result_getitem.data.shape[0] == len(result_getitem.headers)
+    assert result_getitem.data.shape[1] == len(result_getitem.samples)
+
+    # Check that the headers and samples contains proper values
```
a-arefina, line 107:
```suggestion
    # Check that the headers and samples contain proper values
```

seismicpro/src/tests/gather_test.py
```diff
@@ -1,43 +1,160 @@
...
+    ## This is probably not the best way for the equality check..
+    keys = (keys[0], slice(None)) if len(keys) < 2 else keys
+    keys = tuple(to_list(k) if isinstance(k, tuple) else k for k in keys)
```
a-arefina, line 110: It seems that you can swap theese 2 lines,  the result will be the same, but the readability would increase

seismicpro/src/tests/gather_test.py
```diff
@@ -1,43 +1,160 @@
...
+    assert np.allclose(result_getitem.headers, gather.headers.iloc[keys[0]].values)
+    assert np.allclose(result_getitem.samples, gather.samples[keys[1]])
+
+
+@pytest.mark.parametrize('key', [[0, 3, 1]])
```
a-arefina, line 115: you might also want to check that the sort_by is preserved when this is needed

seismicpro/src/gather.py
```diff
@@ -91,19 +91,69 @@ def shape(self):
...
+                new_k = slice(k, k+1) # Switch from simple indexing to a slice to keep array dims
+            elif isinstance(k, tuple):
+                new_k = to_list(k) # Force advanced indexing for `samples`
+            else:
+                new_k = k
+            indices = indices + (new_k, )
+
+        new_self = self.copy(ignore=['data', 'headers', 'samples'])
+        new_self.data = self.data[indices]
+        if new_self.data.ndim != 2:
```
a-arefina, line 143: plz provide an example of this case, a test would be the best!

seismicpro/src/gather.py
```diff
@@ -91,19 +91,69 @@ def shape(self):
         return self.data.shape
 
     def __getitem__(self, key):
-        """Select gather headers by their names.
+        """Either select gather headers by their names or create a new `Gather` with specified traces and samples
```
nikita-klsh, line 94: Please specify explicitly that in case headers - np.array returned, not dataframe(as expected)
```suggestion
        """Either select gather headers values by their names or create a new `Gather` with specified traces and samples
```

seismicpro/src/gather.py
```diff
@@ -91,19 +91,69 @@ def shape(self):
...
         """
-        return self.headers[key].values
+        # If key is string or array of str, return header columns with specified names
+        keys_array = np.array(to_list(key))
+        if keys_array.dtype.type == np.str_:
+            self.validate(required_header_cols=keys_array)
+            # We avoid using pandas indexing with multiple columns to speed up the selection of headers from gathers
+            # with the small number of traces.
+            headers = []
+            for col in keys_array:
```
nikita-klsh, line 122: What do you think about renaming `col` -> `column`. 
Now we can do it  just here, and I will unify namings everywhere in upcoming PR

seismicpro/src/gather.py
```diff
@@ -91,19 +91,69 @@ def shape(self):
...
+                header = self.headers[col] if col in self.headers.columns else self.headers.index.get_level_values(col)
+                headers.append(header.values)
+            return np.column_stack(headers)
+
+        # Another case is to slice gather with the given key. Here key can be any type that might be processed by
+        # np.array's getitem.
+        key = (key, ) if not isinstance(key, tuple) else key
+        key = key + (slice(None), ) if len(key) == 1 else key
+        indices = ()
+        for k in key:
```
nikita-klsh, line 132: Please provide meaningful name for `k`. Is it something like `axis_slice`, right?

seismicpro/src/tests/gather_test.py
```diff
@@ -1,43 +1,160 @@
...
+
+    expected = gather.headers[key].values
+    assert np.allclose(result_getitem, result_get_item)
+    assert result_getitem.shape == (gather.shape[0], len(to_list(key)))
+    assert np.allclose(result_getitem.reshape(-1), expected.reshape(-1))
+    assert result_getitem.dtype == expected.dtype
+
+
+key_types = [0, (0, ), (0, 1), [0], [0, 4], slice(None, 5, None)]
+fail_keys = list(product([(0, 1), (0, ), [0], [0, 4]], repeat=2))
```
nikita-klsh, line 79: Why these are fail keys, if they are valid for gather.data[key]?
Also, `gather[:, 1]` valid, `gather[:, -1]` not valid
By now its unclear, do we fully inherit `np.array.__getitem__` method(as you mention in the comment) or not

seismicpro/src/gather.py
```diff
@@ -91,19 +91,69 @@ def shape(self):
...
+                new_k = k
+            indices = indices + (new_k, )
+
+        new_self = self.copy(ignore=['data', 'headers', 'samples'])
+        new_self.data = self.data[indices]
+        if new_self.data.ndim != 2:
+            raise ValueError("Given `key` leads either to ambiguous behavior when processing samples and headers or to"
+                             " a change in the dimension of the data")
+        if new_self.data.size == 0:
+            raise ValueError("Given `key` results in empty object")
```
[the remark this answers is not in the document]
alexandrkuvaev: I'd raise, it's possible to mess up with gather indexing, resulting in empty data

seismicpro/src/gather.py
alexandrkuvaev: +, probably `axis_indexer` is a bit better

alexandrkuvaev (changes requested):

seismicpro/src/tests/gather_test.py
anton-br: why? Here I try all variants of keys that may be passed into getitem and fail in some of them. I think is a solid part

seismicpro/src/gather.py
anton-br: they are already there
this raise will be called if the `key` has the following structure (array, array).

seismicpro/src/gather.py
anton-br: i would leave it as this, because it perfectly fit the next line :)

seismicpro/src/gather.py
anton-br: this case is already there
The raise will be called if the `key` has a structure like (array, array).

seismicpro/src/tests/gather_test.py
anton-br: why? In this case i provide all possible combinations of key for getitem (expect bool, I forgot about them, maybe I'll add them later). Some of the keys are failing, I think is a normal behaviour

seismicpro/src/tests/gather_test.py
anton-br: > Why these are fail keys, if they are valid for gather.data[key]?

No, they don't. you cannot use `key` like ((0, 1), (0, 1))

> Also, gather[:, 1] valid, gather[:, -1] not valid

good point. Do we want to use negative notation? (It is not valid for now only because of implementation)

> By now its unclear, do we fully inherit np.array.__getitem__ method(as you mention in the comment) or not

We don't follow only the shapes. In our case `data` is always 2d while getitem from numpy may return 1d arrays sometimes.

alexandrkuvaev: > Do we want to use negative notation?

I'd like to, it can be fixed with a one-liner
```
axis_indexer %= corresponding_data_size
```
before converting `axis_indexer` to a `slice`.

alexandrkuvaev (approved):

seismicpro/src/tests/gather_test.py
a-arefina: This is a normal practice. And you actually have 2 very different branches of code in the same function, so it would be better to split them into 2 functions

seismicpro/src/gather.py
```diff
@@ -129,14 +129,15 @@ def __getitem__(self, key):
...
-                new_k = k
-            indices = indices + (new_k, )
+        for axis_indexer, axis_shape in zip(key, self.shape):
+            if isinstance(axis_indexer, (int, np.integer)):
+                # Convert negative array index to a corresponding positive one
+                axis_indexer %= axis_shape
+                # Switch from simple indexing to a slice to keep array dims
+                axis_indexer = slice(axis_indexer, axis_indexer+1)
+            elif isinstance(axis_indexer, tuple):
+                axis_indexer = to_list(axis_indexer) # Force advanced indexing for `samples`
```
a-arefina, line 139:
```suggestion
                axis_indexer = list(axis_indexer) # Force advanced indexing for `samples`
```

seismicpro/src/gather.py
a-arefina: So why do you check this in such an idirect way?

anton-br: this check is much simpler than check `key` tuple that it contains two structures that trigger advanced indexing

seismicpro/src/gather.py
```diff
@@ -91,19 +91,71 @@ def shape(self):
...
+            # We avoid using pandas indexing with multiple columns to speed up the selection of headers from gathers
+            # with the small number of traces.
+            headers = []
+            for col in keys_array:
+                header = self.headers[col] if col in self.headers.columns else self.headers.index.get_level_values(col)
+                headers.append(header.values)
+            return np.column_stack(headers)
+
+        # Another case is to slice gather with the given key. Here key can be any type that might be processed by
+        # np.array's getitem.
```
nikita-klsh, line 128: Please rewrite a bit, that actually not any key valid for `numpy`, valid for `gather`

nikita-klsh (approved):

PR #71, opened 2021-11-25
dump picking

nikita-klsh:
- Add action for fbp dump. 
- Slightly increase fbp load functionality.
- Pass kwargs to @action through @batch_method
- Unify headers dtypes


seismicpro/src/batch.py
```diff
@@ -300,3 +300,40 @@ def split_model_outputs(self, src, dst, shapes):
         else:
             raise ValueError(f"dst must be either `str` or `NamedExpression`, not {type(dst)}.")
         return self
+
+    @action(use_lock=True)
+    def dump_first_breaks(self, path, src, trace_id_columns=('FieldRecord', 'TraceNumber'),
```
alexandrkuvaev, line 305: Why is it a `SeismicBatch` method, not a `Gather` one? Usually, you store picking as a gather header, not a separate component, at least for visualization purposes.

seismicpro/src/batch.py
```diff
@@ -300,3 +300,40 @@ def split_model_outputs(self, src, dst, shapes):
...
+        kwargs : misc, optional
+            Additional keyword arguments to `df.to_string`.
+
+        Returns
+        -------
+        batch : SeismicBatch
+            Batch unchanged 
+        """
+        data = getattr(self, src)
+        df = self.index.headers.reset_index()[list(trace_id_columns)]
```
alexandrkuvaev, line 334: This may fail in the case of survey concatenation/merging. Better use gather headers here.

seismicpro/src/batch.py
```diff
@@ -300,3 +300,40 @@ def split_model_outputs(self, src, dst, shapes):
...
+        df['TMP_FBP'] = np.concatenate(data)
+
+        with open(path, 'a+') as f:
```
alexandrkuvaev, line 337: Why `a+`, not `a`?

seismicpro/src/batch.py
```diff
@@ -300,3 +300,40 @@ def split_model_outputs(self, src, dst, shapes):
...
+            f.write(df.to_string(index=False, header=False, col_space=col_space, decimal=decimal, **kwargs) + '\n')
```
alexandrkuvaev, line 338: Do you really need to redefine defaults for `col_space` and `decimal` or you can pass them as `kwargs`?

alexandrkuvaev (changes requested):

seismicpro/src/batch.py
nikita-klsh: These defaults are required by `flatiron` soft
With these defaults the dump file can be directly loaded to both `survey` and `flatiron` soft

seismicpro/src/survey.py
```diff
@@ -464,19 +473,19 @@ def load_first_breaks(self, path, first_breaks_col='FirstBreak'):
...
         ValueError
-            If `FieldRecord` or `TraceNumber` headers were not loaded.
-            If ('FieldRecord', 'TraceNumber') pairs from the file do not match those in `self.headers`.
+            If not all the `trace_id_columns` are loaded in `self.headers`
+            If there is not a single match of rows from the file with those in `self.headers`.
         """
-        segy_columns = ['FieldRecord', 'TraceNumber']
-        first_breaks_columns = segy_columns + [first_breaks_col]
-        first_breaks_df = pd.read_csv(path, names=first_breaks_columns, delim_whitespace=True, decimal=',')
+        file_columns = trace_id_columns + (first_breaks_column, )
```
anton-br, line 479: feels bad about this line without a cast to any array-like.. for example
```suggestion
        file_columns = to_list(trace_id_columns) + [first_breaks_column]
```

seismicpro/src/gather.py
```diff
@@ -664,6 +664,41 @@ def mask_to_pick(self, threshold=0.5, first_breaks_col="FirstBreak", mask_attr="
...
+
+        Each row in the resulted file corresponds to the first break picking of the trace.
+        All but the last columns stores values from `self.headers[trace_id_columns]`.
+        The last column stores fbp times from batch component.
+
+        Parameters
+        ----------
+        path : str
+            Path to the file.
+        trace_id_columns : tuple of str, defaults to ('FieldRecord', 'TraceNumber')
```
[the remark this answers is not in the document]
anton-br: Also, in your code, you highly rely on the fact that it is a tuple, why? Why don't I use `TRACE_SEQUENCE_FILE` as `trace_id_columns`? It is also a unique trace identifier...

seismicpro/src/gather.py
```diff
@@ -664,6 +664,41 @@ def mask_to_pick(self, threshold=0.5, first_breaks_col="FirstBreak", mask_attr="
         self[first_breaks_col] = convert_mask_to_pick(mask, self.sample_rate, threshold)
         return self
 
+    @batch_method(target='for', use_lock=True)
```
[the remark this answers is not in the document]
nikita-klsh: this action is not within-batch thread-safe to run target=`t`
even if we make it thread-safe it will become ~for loop in terms of performance

seismicpro/src/gather.py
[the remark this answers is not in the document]
nikita-klsh: Sometimes trace_id cols are coords. 
Wrapped it with `to_list` for convenience

seismicpro/src/gather.py
```diff
@@ -664,6 +664,41 @@ def mask_to_pick(self, threshold=0.5, first_breaks_col="FirstBreak", mask_attr="
...
+    def dump_first_breaks(self, path, trace_id_columns=('FieldRecord', 'TraceNumber'),
+                          first_breaks_column='FirstBreak', col_space=8, decimal=',', **kwargs):
+        """ Save first break picking times to the file.
```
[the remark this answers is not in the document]
nikita-klsh: these files actually do not have a special format, the format is described by the file structure

seismicpro/src/gather.py
```diff
@@ -664,6 +664,49 @@ def mask_to_pick(self, threshold=0.5, first_breaks_col="FirstBreak", mask_attr="
...
+        For each row: all but the last columns stores values from `trace_id_columns` headers,
+        the last column stores fbp times from `first_breaks_column` header.
+ 
+        Parameters
+        ----------
+        path : str
+            Path to the file.
+        trace_id_columns : tuple of str, defaults to ('FieldRecord', 'TraceNumber')
+            Columns names from `self.headers` that act as trace id. These would present in the file.
+        first_breaks_column : str, defaults to 'FirstBreak'
```
alexandrkuvaev, line 682: The same arg in `pick_to_mask` and `mask_to_pick` is called `first_breaks_col`, I'd stick to this naming. If so, let's also rename `trace_id_columns` to `trace_id_cols`

seismicpro/src/gather.py
```diff
@@ -664,6 +664,49 @@ def mask_to_pick(self, threshold=0.5, first_breaks_col="FirstBreak", mask_attr="
...
+            Column name from `self.headers` where fbp is stored.
+        col_space : int, defaults to 8
+            The minimum width of each column.
+        decimal : str, default to ','
```
alexandrkuvaev, line 686: We don't use `pandas` to create the string to dump now, so maybe we should consider renaming it to something more clear, indicating that this is a separator. Maybe `decimal_sep`.

seismicpro/src/gather.py
```diff
@@ -664,6 +664,49 @@ def mask_to_pick(self, threshold=0.5, first_breaks_col="FirstBreak", mask_attr="
...
+        else:
+            fmt = ['{:{col_space}.0f}'] * (rows.shape[1] - 1) + ['{:{col_space}.2f}']
+        fmt = ''.join(fmt)
+
+        with open(path, 'a') as f:  #pylint: disable=unspecified-encoding
+            for row in rows:
+                string = fmt.format(*row, col_space=col_space)
+                if decimal != '.':
+                    string = string.replace('.', decimal, 1)
+                f.write(string + '\n')
```
alexandrkuvaev, line 708: Won't it be faster to format all rows into a single string and then write it in a single `write`?

seismicpro/src/gather.py
```diff
@@ -664,6 +664,49 @@ def mask_to_pick(self, threshold=0.5, first_breaks_col="FirstBreak", mask_attr="
...
+        rows = self[to_list(trace_id_columns) + [first_breaks_column]]
+
+        # segy spec states that all headers values are integers, first break values tho can be float
+        if rows.dtype == int:
+            fmt = ['{:{col_space}.0d}'] * rows.shape[1]
+        else:
+            fmt = ['{:{col_space}.0f}'] * (rows.shape[1] - 1) + ['{:{col_space}.2f}']
+        fmt = ''.join(fmt)
+
+        with open(path, 'a') as f:  #pylint: disable=unspecified-encoding
```
alexandrkuvaev, line 703: Let's pass `encoding` directly with a default of `"UTF-8"` as we do in all other pure-python load/dump functions

seismicpro/src/gather.py
```diff
@@ -664,6 +664,49 @@ def mask_to_pick(self, threshold=0.5, first_breaks_col="FirstBreak", mask_attr="
...
+        Returns
+        -------
+        self : Gather
+            Gather unchanged
+        """
+        rows = self[to_list(trace_id_columns) + [first_breaks_column]]
+
+        # segy spec states that all headers values are integers, first break values tho can be float
+        if rows.dtype == int:
+            fmt = ['{:{col_space}.0d}'] * rows.shape[1]
```
alexandrkuvaev, line 698: Precision is not allowed for integers, why not always use the format from the `else` branch?

seismicpro/src/survey.py
```diff
@@ -441,20 +441,27 @@ def load_trace(self, buf, index, limits, trace_length):
...
+        the trace from the file with the corresponding trace in `self.headers`.
 
         Parameters
         ----------
         path : str
             A path to the file with first break times in milliseconds.
-        first_breaks_col : str, optional, defaults to 'FirstBreak'
+        trace_id_columns : tuple of str, defaults to ('FieldRecord', 'TraceNumber')
+            All but the last columns names in the file.
+        first_breaks_column : str, optional, defaults to 'FirstBreak'
```
alexandrkuvaev, line 459: Similar comment about variable naming

seismicpro/src/survey.py
```diff
@@ -464,19 +471,17 @@ def load_first_breaks(self, path, first_breaks_col='FirstBreak'):
...
-        first_breaks_columns = segy_columns + [first_breaks_col]
-        first_breaks_df = pd.read_csv(path, names=first_breaks_columns, delim_whitespace=True, decimal=',')
+        file_columns = to_list(trace_id_columns) + [first_breaks_column]
+        first_breaks_df = pd.read_csv(path, names=file_columns, delim_whitespace=delim_whitespace, **kwargs)
 
-        headers = self.headers.reset_index()
-        missing_cols = set(segy_columns) - set(headers)
-        if missing_cols:
-            raise ValueError(f'Missing {missing_cols} column(s) required for first break loading.')
+        # little insurance: if we messed with decimal, dtype still be infered
```
alexandrkuvaev, line 479: Can't you force it using a `dtype` dict in `read_csv`?

alexandrkuvaev (changes requested):

seismicpro/src/survey.py
```diff
@@ -464,19 +473,17 @@ def load_first_breaks(self, path, first_breaks_col='FirstBreak'):
...
+        if first_breaks_df[first_breaks_column].dtype == object:
+            first_breaks_df[first_breaks_column] = pd.to_numeric(first_breaks_df[first_breaks_column])
 
-        headers = headers.merge(first_breaks_df, on=segy_columns)
+        headers = self.headers.reset_index()
+        headers = headers.merge(first_breaks_df, on=trace_id_columns)
```
anton-br, line 486: You have deleted one of the raises, please check that it fails with the same error

seismicpro/src/gather.py
anton-br, line 686: unify default in docs with real default value

seismicpro/src/decorators.py
```diff
@@ -185,6 +190,8 @@ def method(self, index, *args, src=None, dst=None, **kwargs):
                             obj_arguments.arguments[arg_name] = getattr(self, arg_val)[pos]
                 getattr(self, dst)[pos] = obj_method(*obj_arguments.args, **obj_arguments.kwargs)
             method.__name__ = method_name
+            if action_params[method_name]:
+                return action(**action_params[method_name])(apply_to_each_component(method))
             return action(apply_to_each_component(method))
```
anton-br, line 195: instead `action_params[method_name] = getattr(method, 'action_params')` write `action_params[method_name] = getattr(method, 'action_params', {})`

and return `action(**action_params[method_name])(apply_to_each_component(method))`

seismicpro/src/gather.py
nikita-klsh: ~10% faster: 10ms vs 11 ms for 3k traces gather.
Readability and consistency tho suffer a bit (pandas itself and numpy dumps its objects in row-wise manner).

seismicpro/src/gather.py
nikita-klsh: ++ integer format is incorrect now.

We can always use `else` branch, but this requires implicit cast `rows` to `float`(`rows.dtype == int` if `fbp.dtype == int`(picking  comes from the model pipeline))

So its either complicated formatted or unnecessary cast

seismicpro/src/gather.py
alexandrkuvaev: The code will be even shorter and more concise than it is now

seismicpro/src/gather.py
```diff
@@ -664,6 +664,42 @@ def mask_to_pick(self, threshold=0.5, first_breaks_col="FirstBreak", mask_attr="
...
+        Returns
+        -------
+        self : Gather
+            Gather unchanged
+        """
+        rows = self[to_list(trace_id_columns) + [first_breaks_col]]
+
+        with open(path, 'a', encoding=encoding) as f:
+            # segy spec states that all headers values are integers, first break values tho can be float
+            fmt_for_row = ['{:{col_space}.0f}'] * (rows.shape[1] - 1) + ['{:{col_space}.2f}'] + ['\n']
```
alexandrkuvaev, line 697:
```suggestion
            row_fmt = '{:{col_space}.0f}' * (rows.shape[1] - 1) + '{:{col_space}.2f}\n'
```

seismicpro/src/gather.py
```diff
@@ -664,6 +664,42 @@ def mask_to_pick(self, threshold=0.5, first_breaks_col="FirstBreak", mask_attr="
...
+            fmt_for_rows = ''.join(fmt_for_row) * len(rows)
```
alexandrkuvaev, line 698:
```suggestion
            fmt = row_fmt * len(rows)
```

seismicpro/src/gather.py
```diff
@@ -664,6 +664,42 @@ def mask_to_pick(self, threshold=0.5, first_breaks_col="FirstBreak", mask_attr="
...
+
+        Each row in the resulted file corresponds to the first break picking of the trace.
+        For each row: all but the last columns stores values from `trace_id_columns` headers,
+        the last column stores fbp times from `first_breaks_col` header.
+
+        Parameters
+        ----------
+        path : str
+            Path to the file.
+        trace_id_columns : tuple of str, defaults to ('FieldRecord', 'TraceNumber')
```
alexandrkuvaev, line 680: Let's call it `trace_id_cols` to match naming in `load_first_breaks` and other args

seismicpro/src/survey.py
```diff
@@ -132,10 +132,12 @@ def __init__(self, path, header_index, header_cols=None, name=None, limits=None,
         for column in load_headers:
             headers[column] = self.segy_handler.attributes(segyio.tracefield.keys[column])[:]
 
-        headers = pd.DataFrame(headers)
+        # According to segy spec, headers values at most 4 byte integers
+        headers = pd.DataFrame(headers, dtype=np.int32)
         # TRACE_SEQUENCE_FILE is reconstructed manually since it can be omitted according to the SEG-Y standard
         # but we rely on it during gather loading.
-        headers["TRACE_SEQUENCE_FILE"] = np.arange(1, self.segy_handler.tracecount+1)
+        tsf_dtype = np.int32 if len(headers) < np.iinfo(np.int32).max else np.int64
```
alexandrkuvaev, line 139: Try using `np.min_scalar_type` for `len(headers)` here

seismicpro/src/survey.py
```diff
@@ -464,19 +479,24 @@ def load_first_breaks(self, path, first_breaks_col='FirstBreak'):
...
-            If `FieldRecord` or `TraceNumber` headers were not loaded.
-            If ('FieldRecord', 'TraceNumber') pairs from the file do not match those in `self.headers`.
+            If there is not a single match of rows from the file with those in `self.headers`.
         """
-        segy_columns = ['FieldRecord', 'TraceNumber']
-        first_breaks_columns = segy_columns + [first_breaks_col]
-        first_breaks_df = pd.read_csv(path, names=first_breaks_columns, delim_whitespace=True, decimal=',')
+        # if decimal is not provided, try to infer it from the first line
+        if decimal is None:
+            encoding=kwargs.get('encoding', 'UTF-8')
```
alexandrkuvaev, line 486: Add it to args directly and also pass to `read_csv`

alexandrkuvaev (changes requested):

seismicpro/src/gather.py
```diff
@@ -664,6 +664,43 @@ def mask_to_pick(self, threshold=0.5, first_breaks_col="FirstBreak", mask_attr="
...
         return self
 
+    @batch_method(target='for', use_lock=True)
+    def dump_first_breaks(self, path, trace_id_cols=('FieldRecord', 'TraceNumber'), first_breaks_col='FirstBreak',
+                          col_space=8, encoding="UTF-8"):
+        """ Save first break picking times to the file.
+
+        Each row in the resulted file corresponds to the first break picking of the trace.
+        For each row: all but the last columns stores values from `trace_id_columns` headers,
+        the last column stores fbp times from `first_breaks_col` header.
```
a-arefina, line 674:
```suggestion
        Each row in the resulting file corresponds to the first break picking of the trace.
        For each row: all but the last columns store values from `trace_id_columns` headers,
        the last column stores first break picking times from `first_breaks_col` header.
```

seismicpro/src/gather.py
```diff
@@ -664,6 +664,43 @@ def mask_to_pick(self, threshold=0.5, first_breaks_col="FirstBreak", mask_attr="
...
+
+        Parameters
+        ----------
+        path : str
+            Path to the file.
+        trace_id_cols : tuple of str, defaults to ('FieldRecord', 'TraceNumber')
+            Columns names from `self.headers` that act as trace id. These would present in the file.
```
a-arefina, line 681:
```suggestion
            Columns names from `self.headers` that act as trace id. These would be present in the file.
```

seismicpro/src/gather.py
```diff
@@ -664,6 +664,43 @@ def mask_to_pick(self, threshold=0.5, first_breaks_col="FirstBreak", mask_attr="
...
+        first_breaks_col : str, defaults to 'FirstBreak'
+            Column name from `self.headers` where fbp is stored.
```
a-arefina, line 683:
```suggestion
            Column name from `self.headers` where first break picking times are stored.
```

seismicpro/src/gather.py
```diff
@@ -664,6 +664,43 @@ def mask_to_pick(self, threshold=0.5, first_breaks_col="FirstBreak", mask_attr="
...
+        encoding : str, optional, defaults to "UTF-8"
+            File encoding.
+        Returns
+        -------
+        self : Gather
+            Gather unchanged
+        """
+        rows = self[to_list(trace_id_cols) + [first_breaks_col]]
+
+        # segy spec states that all headers values are integers, first break values tho can be float
```
a-arefina, line 695:
```suggestion
        # SEG-Y specification states that all headers values are integers, but first break values can be float
```

seismicpro/src/survey.py
```diff
@@ -441,20 +443,35 @@ def load_trace(self, buf, index, limits, trace_length):
         """
         return self.segy_handler.xfd.gettr(buf, index, 1, 1, limits.start, limits.stop, limits.step, trace_length)
 
-    def load_first_breaks(self, path, first_breaks_col='FirstBreak'):
-        """Load times of first breaks and save them to `headers`.
+    def load_first_breaks(self, path, trace_id_cols = ('FieldRecord', 'TraceNumber'), first_breaks_col='FirstBreak',
+                          delim_whitespace=True, decimal=None, encoding="UTF-8", **kwargs):
+        """Load first break picking times from the file and save them to the new column in headers.
```
a-arefina, line 448:
```suggestion
        """Load first break picking times from a file and save them to a new column in headers.
```

seismicpro/src/survey.py
```diff
@@ -441,20 +443,35 @@ def load_trace(self, buf, index, limits, trace_length):
...
         Parameters
         ----------
         path : str
             A path to the file with first break times in milliseconds.
+        trace_id_cols : tuple of str, defaults to ('FieldRecord', 'TraceNumber')
+            All but the last columns names in the file.
         first_breaks_col : str, optional, defaults to 'FirstBreak'
             Column name in `self.headers` where loaded first break times will be stored.
+        delim_whitespace: bool, defaults to True
+            Specifies whether or not whitespace will be used as the sep. See `pd.read_csv` for more details.
```
a-arefina, line 467:
```suggestion
            Specifies whether or not whitespace will be used as a delimiter. See `pd.read_csv` for more details.
```

seismicpro/src/survey.py
```diff
@@ -441,20 +443,35 @@ def load_trace(self, buf, index, limits, trace_length):
         """
         return self.segy_handler.xfd.gettr(buf, index, 1, 1, limits.start, limits.stop, limits.step, trace_length)
 
-    def load_first_breaks(self, path, first_breaks_col='FirstBreak'):
-        """Load times of first breaks and save them to `headers`.
+    def load_first_breaks(self, path, trace_id_cols = ('FieldRecord', 'TraceNumber'), first_breaks_col='FirstBreak',
```
a-arefina, line 446: What about adding `inplace` parameter since this method can destroy headers?

a-arefina (commented):
Most comments suggest minor fixes in docs. Please avoid abbreviations

seismicpro/src/gather.py
alexandrkuvaev: Also `trace_id_columns` -> `trace_id_cols`

seismicpro/src/gather.py
```diff
@@ -664,6 +664,43 @@ def mask_to_pick(self, threshold=0.5, first_breaks_col="FirstBreak", mask_attr="
...
+            Path to the file.
+        trace_id_cols : tuple of str, defaults to ('FieldRecord', 'TraceNumber')
+            Columns names from `self.headers` that act as trace id. These would present in the file.
+        first_breaks_col : str, defaults to 'FirstBreak'
+            Column name from `self.headers` where fbp is stored.
+        col_space : int, defaults to 8
+            The minimum width of each column.
+        encoding : str, optional, defaults to "UTF-8"
+            File encoding.
+        Returns
```
alexandrkuvaev, line 688: Add an empty line before returns block

seismicpro/src/survey.py
```diff
@@ -441,20 +443,35 @@ def load_trace(self, buf, index, limits, trace_length):
...
 
         Parameters
         ----------
         path : str
             A path to the file with first break times in milliseconds.
+        trace_id_cols : tuple of str, defaults to ('FieldRecord', 'TraceNumber')
+            All but the last columns names in the file.
         first_breaks_col : str, optional, defaults to 'FirstBreak'
             Column name in `self.headers` where loaded first break times will be stored.
+        delim_whitespace: bool, defaults to True
```
alexandrkuvaev, line 466: Let's change it to a `delimiter` arg with default `\s+`

seismicpro/src/survey.py
```diff
@@ -464,19 +479,24 @@ def load_first_breaks(self, path, first_breaks_col='FirstBreak'):
...
-        first_breaks_df = pd.read_csv(path, names=first_breaks_columns, delim_whitespace=True, decimal=',')
+        # if decimal is not provided, try to infer it from the first line
+        if decimal is None:
+            encoding=kwargs.get('encoding', 'UTF-8')
+            with open(path, 'r', encoding=encoding) as f:
+                row = f.readline()
+                if not ' ' in row:  # coma-separated-values
+                    decimal = '.'
+                else:               # fixed-width-column
+                    decimal = ',' if ',' in row else '.'
```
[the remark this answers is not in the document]
alexandrkuvaev:
```suggestion
            with open(path, 'r', encoding=encoding) as f:
                row = f.readline()
            decimal = '.' if '.' in row else ','
```

alexandrkuvaev (changes requested):

seismicpro/src/survey.py
```diff
@@ -441,20 +443,38 @@ def load_trace(self, buf, index, limits, trace_length):
...
+
+        The file can have any format that can be read by `pd.read_csv`,
+        by default it's expected to have whitespace separated values, notice `delim_whitespace=True`.
 
         Parameters
         ----------
         path : str
             A path to the file with first break times in milliseconds.
+        trace_id_cols : tuple of str, defaults to ('FieldRecord', 'TraceNumber')
+            All but the last columns names in the file.
```
alexandrkuvaev, line 464:
```suggestion
            Headers, whose values are stored in all but the last columns of the file.
```

seismicpro/src/survey.py
```diff
@@ -441,20 +443,38 @@ def load_trace(self, buf, index, limits, trace_length):
...
         first_breaks_col : str, optional, defaults to 'FirstBreak'
             Column name in `self.headers` where loaded first break times will be stored.
+        delimiter: str, defaults to '\s+'
+            Delimiter to use. See `pd.read_csv` for more details.
+        decimal : str, defaults to None
+            Character to recognize as decimal point.
+            In case None tries to infer decimal from the first line of the file.
```
alexandrkuvaev, line 471:
```suggestion
            If `None`, it is inferred from the first line of the file.
```

seismicpro/src/survey.py
```diff
@@ -441,20 +443,38 @@ def load_trace(self, buf, index, limits, trace_length):
...
+        encoding : str, optional, defaults to "UTF-8"
+            File encoding.
+        inplace : bool, optional, defaults to False
+            Whether to load first break times inplace or to a survey copy.
+        kwargs : misc, optional
+            Additional keyword arguments to pass to  `pd.read_csv`.
```
alexandrkuvaev, line 477:
```suggestion
            Additional keyword arguments to pass to `pd.read_csv`.
```

alexandrkuvaev (changes requested):

a-arefina (approved):

seismicpro/src/survey.py
```diff
@@ -441,20 +444,37 @@ def load_trace(self, buf, index, limits, trace_length):
         """
         return self.segy_handler.xfd.gettr(buf, index, 1, 1, limits.start, limits.stop, limits.step, trace_length)
 
-    def load_first_breaks(self, path, first_breaks_col='FirstBreak'):
-        """Load times of first breaks and save them to `headers`.
+    # pylint: disable=anomalous-backslash-in-string
+    def load_first_breaks(self, path, trace_id_cols=('FieldRecord', 'TraceNumber'), first_breaks_col='FirstBreak',
+                          delimiter='\s+', decimal=None, encoding="UTF-8", inplace=False, **kwargs):
+        """Load first break picking times from a file and save them to a new column in headers.
```
alexandrkuvaev, line 450:
```suggestion
        """Load times of first breaks from a file and save them to a new column in headers.
```

alexandrkuvaev (changes requested):

alexandrkuvaev (approved):

seismicpro/src/gather.py
```diff
@@ -664,6 +664,44 @@ def mask_to_pick(self, threshold=0.5, first_breaks_col="FirstBreak", mask_attr="
...
         return self
 
+    @batch_method(target='for', use_lock=True)
+    def dump_first_breaks(self, path, trace_id_cols=('FieldRecord', 'TraceNumber'), first_breaks_col='FirstBreak',
+                          col_space=8, encoding="UTF-8"):
+        """ Save first break picking times to the file.
+
+        Each row in the resulting file corresponds to the first break picking of the trace.
+        For each row: all but the last columns store values from `trace_id_cols` headers,
+        the last column stores first break times from `first_breaks_col` header.
```
dpodvyaznikov, line 674:
```suggestion
        Each line in the resulting file corresponds to one trace, where all columns but the last one store values from `trace_id_cols` headers and identify the trace while the last column stores first break time from `first_breaks_col` header.
```

seismicpro/src/gather.py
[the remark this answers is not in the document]
dpodvyaznikov:
```suggestion
        """ Save first break picking times to a file.
```

dpodvyaznikov (changes requested):

dpodvyaznikov (approved):

PR #73, opened 2021-12-07
mark dead traces

a-arefina:
keeping dead trace indicator in headers allows convenient fitering

### DIFF
--- a/seismicpro/src/survey.py
+++ b/seismicpro/src/survey.py
@@ -224,9 +224,9 @@ class Survey:  # pylint: disable=too-many-instance-attributes
 
         After the method is executed `has_stats` flag is set to `True` and all the calculated values can be obtained
         via corresponding attributes.
-        
-        The method also appends 'DeadTrace' column to `headers`. The value of this column is 1 if the corresponding trace 
-        is constant (dead), otherwise, the value is 0.
+
+        The method also appends 'DeadTrace' column to `headers`.
+        The value of this column is 1 if the corresponding trace is constant (dead), otherwise, the value is 0.
 
         Parameters
         ----------
@@ -287,7 +287,7 @@ class Survey:  # pylint: disable=too-many-instance-attributes
             # Sample random traces to calculate approximate quantiles
             if i < n_quantile_traces:
                 traces_buf[i] = trace
-                
+
         self.n_dead_traces = len(dead_indices)
         self.headers['DeadTrace'] = 0
         self.headers.loc[self.headers["TRACE_SEQUENCE_FILE"].isin(dead_indices), 'DeadTrace'] = 1
@@ -683,7 +683,7 @@ class Survey:  # pylint: disable=too-many-instance-attributes
         if limits[-1] < 0:
             raise ValueError('Negative step is not allowed.')
         return slice(*limits)
-    
+
     def remove_dead_traces(self, recollect_stats=False, inplace=False):
         """ Removes dead (constant) traces from survey's data.
         Recalculates surveys statistics if needed
--- a/seismicpro/src/survey.py
+++ b/seismicpro/src/survey.py
@@ -266,6 +266,7 @@ class Survey:  # pylint: disable=too-many-instance-attributes
 
         traces_buf = np.empty((n_quantile_traces, self.samples_length), dtype=np.float32)
         trace = np.empty(self.samples_length, dtype=np.float32)
+        dead_indices = []
 
         # Accumulate min, max, mean and std values of survey traces
         for i, pos in tqdm(enumerate(traces_pos), desc=f"Calculating statistics for survey {self.name}",
@@ -277,11 +278,17 @@ class Survey:  # pylint: disable=too-many-instance-attributes
             global_sum += trace_sum
             global_sq_sum += trace_sq_sum
             traces_length += len(trace)
-            self.n_dead_traces += np.isclose(trace_min, trace_max)
+
+            if np.isclose(trace_min, trace_max):
+                dead_indices.append(pos + 1)
 
             # Sample random traces to calculate approximate quantiles
             if i < n_quantile_traces:
                 traces_buf[i] = trace
+                
+        self.n_dead_traces = len(dead_indices)
+        self.headers['DeadTrace'] = 0
+        self.headers.loc[self.headers["TRACE_SEQUENCE_FILE"].isin(dead_indices), 'DeadTrace'] = 1
 
         self.min = np.float32(global_min)
         self.max = np.float32(global_max)
--- a/seismicpro/src/survey.py
+++ b/seismicpro/src/survey.py
@@ -224,6 +224,9 @@ class Survey:  # pylint: disable=too-many-instance-attributes
 
         After the method is executed `has_stats` flag is set to `True` and all the calculated values can be obtained
         via corresponding attributes.
+        
+        The method also appends 'DeadTrace' column to `headers`. The value of this column is 1 if the corresponding trace 
+        is constant (dead), otherwise, the value is 0.
 
         Parameters
         ----------
@@ -262,7 +265,6 @@ class Survey:  # pylint: disable=too-many-instance-attributes
         global_min, global_max = np.inf, -np.inf
         global_sum, global_sq_sum = 0, 0
         traces_length = 0
-        self.n_dead_traces = 0
 
         traces_buf = np.empty((n_quantile_traces, self.samples_length), dtype=np.float32)
         trace = np.empty(self.samples_length, dtype=np.float32)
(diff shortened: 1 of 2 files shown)


seismicpro/src/survey.py
```diff
@@ -277,11 +278,17 @@ def collect_stats(self, indices=None, n_quantile_traces=100000, quantile_precisi
...
-            self.n_dead_traces += np.isclose(trace_min, trace_max)
+
+            if np.isclose(trace_min, trace_max):
+                dead_indices.append(pos + 1)
 
             # Sample random traces to calculate approximate quantiles
             if i < n_quantile_traces:
                 traces_buf[i] = trace
+                
+        self.n_dead_traces = len(dead_indices)
```
alexandrkuvaev, line 289: Remove `self.n_dead_traces = 0` line above then

seismicpro/src/survey.py
```diff
@@ -277,11 +278,17 @@ def collect_stats(self, indices=None, n_quantile_traces=100000, quantile_precisi
...
+        self.headers['DeadTrace'] = 0
+        self.headers.loc[self.headers["TRACE_SEQUENCE_FILE"].isin(dead_indices), 'DeadTrace'] = 1
```
alexandrkuvaev, line 291: You know, that `i` in the above `for` loop represents the index of the processed trace. Thus you can store it in `dead_indices` instead of `pos + 1` to use `iloc` here, which is way faster

seismicpro/src/survey.py
```diff
@@ -277,11 +278,17 @@ def collect_stats(self, indices=None, n_quantile_traces=100000, quantile_precisi
...
+
+            if np.isclose(trace_min, trace_max):
+                dead_indices.append(pos + 1)
 
             # Sample random traces to calculate approximate quantiles
             if i < n_quantile_traces:
                 traces_buf[i] = trace
+                
+        self.n_dead_traces = len(dead_indices)
+        self.headers['DeadTrace'] = 0
```
alexandrkuvaev, line 290: Don't forget to mention this new header in docs

alexandrkuvaev (changes requested):

alexandrkuvaev:
Also think of a `remove_dead_traces` method of a `Survey`, which will run `collect_stats` if they are not present and perform filtering

seismicpro/src/survey.py
a-arefina: but `i` is an index in a shuffled array `praces_pos`, it has no connection with the original traces indices in headers :(

a-arefina:
> Also think of a `remove_dead_traces` method of a `Survey`, which will run `collect_stats` if they are not present and perform filtering

I've added a method for removing dead traces but there are some questions to concider

1) Do we need an option to skip quantiles calculation when running `collect_stats` for performance reasons?
2) deleting dead traces changes survey's stats. I've added a flag to recalculate it, but to do it fairly, we need to save parameters of previous calculation process (indices, n_quantile_traces, quantile_precision, stats_limits). In seismiqb they don't include dead traces in the pool for calculating stats, which removes the need for recalculation after dead traces are deleted. 
3) do we need to recalculate stats in `set_limits`? Maybe there are other methods that can chance stats?

seismicpro/src/survey.py
```diff
@@ -675,6 +684,34 @@ def _process_limits(self, limits):
...
+            Whether to remove data inplace or return a new survey instance, by default False
+
+        Returns
+        -------
+        Survey
+            Survey with no dead traces.
+        """
+        self = maybe_copy(self, inplace)  # pylint: disable=self-cls-assignment
+        if not self.has_stats:
+            self.collect_stats()
```
alexandrkuvaev, line 705: Let's pass all the `kwargs` here

seismicpro/src/survey.py
```diff
@@ -675,6 +684,34 @@ def _process_limits(self, limits):
...
+
+        self.filter(lambda c: c==0, cols='DeadTrace', inplace=True)
+        self.n_dead_traces = 0
+        if recollect_stats:
```
alexandrkuvaev, line 709: I'd change `collect_stats` to use only non-dead traces for stats calculation (which seems pretty reasonable) and remove this check at all

alexandrkuvaev (changes requested):

alexandrkuvaev:
> 1. Do we need an option to skip quantiles calculation when running `collect_stats` for performance reasons?

If you pass `kwargs` as I suggest above, you can do it by passing `n_quantile_traces=0`. But quantile calculation is not generally a bottleneck at least when `n_quantile_traces` is not that great.

> 2. deleting dead traces changes survey's stats. I've added a flag to recalculate it, but to do it fairly, we need to save parameters of previous calculation process (indices, n_quantile_traces, quantile_precision, stats_limits). In seismiqb they don't include dead traces in the pool for calculating stats, which removes the need for recalculation after dead traces are deleted.

I'd calculate stats only for non-dead traces.

> 3. do we need to recalculate stats in `set_limits`? Maybe there are other methods that can chance stats?

I've checked the `Survey`, the following methods can possibly change stats:
* `load_first_breaks`
* `filter`
* `set_limits`
* `generate_supergathers`
All of them change either `headers` or `limits`. One possible way to handle it is to introduce a new flag `dirty_stats`, similar to the one in `VelocityCube`, and check it every time the stats are requested, but I'd keep it as is for now since the fix is not that simple.

seismicpro/src/survey.py
```diff
@@ -675,6 +691,31 @@ def _process_limits(self, limits):
...
         return slice(*limits)
 
+    def remove_dead_traces(self, inplace=False, **kwargs):
+        """ Removes dead (constant) traces from survey's data.
+        Calls `collect_stats` if `self.has_stats` flag is not set
+
+        Parameters
+        ----------
+        inplace : bool, optional
+            Whether to remove data inplace or return a new survey instance, by default False
```
alexandrkuvaev, line 701:
```suggestion
            Whether to remove traces inplace or return a new survey instance, by default False
```

seismicpro/src/survey.py
```diff
@@ -675,6 +691,31 @@ def _process_limits(self, limits):
...
+        Returns
+        -------
+        Survey
+            Survey with no dead traces.
+        """
+        self = maybe_copy(self, inplace)  # pylint: disable=self-cls-assignment
+        if not self.has_stats:
+            self.collect_stats(**kwargs)
+
+        self.filter(lambda dt: dt == False, cols='DeadTrace', inplace=True)
```
alexandrkuvaev, line 714:
```suggestion
        self.filter(lambda dt: ~dt, cols='DeadTrace', inplace=True)
```

seismicpro/src/survey.py
```diff
@@ -267,21 +271,29 @@ def collect_stats(self, indices=None, n_quantile_traces=100000, quantile_precisi
         traces_buf = np.empty((n_quantile_traces, self.samples_length), dtype=np.float32)
         trace = np.empty(self.samples_length, dtype=np.float32)
 
+        dead_indices = []
+        counter = 0
```
alexandrkuvaev, line 275: I'd rename it to something like `cur_quantile_trace_index`

seismicpro/src/survey.py
```diff
@@ -675,6 +691,31 @@ def _process_limits(self, limits):
             raise ValueError('Negative step is not allowed.')
         return slice(*limits)
 
+    def remove_dead_traces(self, inplace=False, **kwargs):
+        """ Removes dead (constant) traces from survey's data.
+        Calls `collect_stats` if `self.has_stats` flag is not set
```
alexandrkuvaev, line 696:
```suggestion
        Calls `collect_stats` if `self.has_stats` flag is not set.
```

alexandrkuvaev (changes requested):

seismicpro/src/survey.py
```diff
@@ -267,21 +271,29 @@ def collect_stats(self, indices=None, n_quantile_traces=100000, quantile_precisi
...
         trace = np.empty(self.samples_length, dtype=np.float32)
 
+        dead_indices = []
+        counter = 0
         # Accumulate min, max, mean and std values of survey traces
-        for i, pos in tqdm(enumerate(traces_pos), desc=f"Calculating statistics for survey {self.name}",
-                           total=len(traces_pos), disable=not bar):
-            self.load_trace(buf=trace, index=pos, limits=limits, trace_length=self.samples_length)
+        for i in tqdm(shuffled_indices, desc=f"Calculating statistics for survey {self.name}",
+                           total=num_traces, disable=not bar):
```
a-arefina, line 278:
```suggestion
        for i in tqdm(shuffled_indices, desc=f"Calculating statistics for survey {self.name}",
                      total=num_traces, disable=not bar):
```

seismicpro/src/survey.py
a-arefina: what about `quantile_traces_counter`?

alexandrkuvaev: Up to you, much better than simply `counter` anyway)

alexandrkuvaev (approved):

anton-br (approved):

dpodvyaznikov (approved):

PR #80, opened 2022-01-18
Add "hist" mode in `gather.plot`

dpodvyaznikov:
Changes summary:
* `gather.plot` with `'hist'` mode can plot histograms of gather data amplitudes or gather's header values;
* now `gather.plot` has `x_tick_src` \ `y_tick_src` parameters that indicate the data source and default label for said axes; `'label'` key of `x\y_ticker` parameter now only controls axis label and can overwrite default provided by `x\y_tick_src`;
* text formatting was moved from `@plotter` decorator to the body of corresponding plot methods, since only plot methods know which of their arguments should be treated as text parameters.
* `None` value is now allowed for `tick_labels` in `set_ticks` function; in this case, all the formatting and locating options (except for `step_labels`, for obvious reasons) are applied to default ticks;

Also, fixed of a small bug in `batch.plot`.

### DIFF
--- a/seismicpro/src/batch.py
+++ b/seismicpro/src/batch.py
@@ -487,12 +487,12 @@ class SeismicBatch(Batch):
 
                 # Format subplot title
                 if title_template is not None:
-                    title = as_dict(title_template, key='label')
-                    label = title.pop("label")
+                    src_title = as_dict(title_template, key='label')
+                    label = src_title.pop("label")
                     format_names = {name for _, name, _, _ in Formatter().parse(label) if name is not None}
-                    format_kwargs = {name: title.pop(name) for name in format_names if name in title}
-                    title["label"] = label.format(src=src, index=index, **format_kwargs)
-                    kwargs["title"] = title
+                    format_kwargs = {name: src_title.pop(name) for name in format_names if name in src_title}
+                    src_title["label"] = label.format(src=src, index=index, **format_kwargs)
+                    kwargs["title"] = src_title
 
                 # Create subplotter config
                 subplot_config = {
--- a/seismicpro/src/decorators.py
+++ b/seismicpro/src/decorators.py
@@ -25,9 +25,7 @@ def plotter(figsize, args_to_unpack=None):
     """Expand the functionality of a plotting method by defining figure creation and saving.
 
     The decorated method is supposed to accept an `ax` argument. If it's not passed during the call, the decorator
-    creates it with the `figsize` provided. Before calling the decorated method, `kwargs` are processed as follows: all
-    text formatting arguments are popped from `kwargs` and set as defaults for 'title', 'x_ticker' and 'y_ticker'.
-
+    creates it with the `figsize` provided.
     A new argument is added for the decorated method:
     save_to : str or dict, optional, defaults to None
         If `str`, a path to save the figure to.
--- a/seismicpro/src/decorators.py
+++ b/seismicpro/src/decorators.py
@@ -6,7 +6,7 @@ from collections import defaultdict
 
 import matplotlib.pyplot as plt
 
-from .utils import to_list, save_figure, set_text_formatting, as_dict
+from .utils import to_list, save_figure, as_dict
 from ..batchflow import action, inbatch_parallel
 
 
@@ -57,7 +57,6 @@ def plotter(figsize, args_to_unpack=None):
     def decorator(method):
         @wraps(method)
         def plot(*args, **kwargs):
-            kwargs = set_text_formatting(kwargs)
             if "ax" in kwargs:
                 return method(*args, **kwargs)
             fig, ax = plt.subplots(1, 1, figsize=kwargs.pop("figsize", figsize))
--- a/seismicpro/src/gather.py
+++ b/seismicpro/src/gather.py
@@ -1042,8 +1042,7 @@ class Gather:
 
     @plotter(figsize=(10, 7))
     def plot(self, mode="seismogram", title=None, x_ticker=None, y_ticker=None, ax=None, **kwargs):
-        """ TODO
-        Plot gather traces.
+        """Plot gather traces.
 
         The traces can be displayed in a number of representations, depending on the `mode` provided. Currently, the
         following options are supported:
--- a/seismicpro/src/gather.py
+++ b/seismicpro/src/gather.py
@@ -1157,6 +1157,7 @@ class Gather:
             ax.set_yscale("log")
         return ax
 
+    # pylint: disable=too-many-arguments
     def _plot_seismogram(self, ax, colorbar=False, qvmin=0.1, qvmax=0.9, x_ticker=None, y_ticker=None,
                          x_tick_src=None, y_tick_src='time', event_headers=None, top_header=None, **kwargs):
         """Plot the gather as a 2d grayscale image of seismic traces."""
--- a/seismicpro/src/gather.py
+++ b/seismicpro/src/gather.py
@@ -1058,12 +1058,12 @@ class Gather:
             * `color`: defines a color for each trace. If a single color is given, it is applied to all the traces
               (defaults to black).
         - `hist`: a histogram of the trace data amplitudes or header values. The mode supports following `kwargs`:
-            * `bins`: number of equal-width bins if an integer; bin edges, including the left edge of the first bin and
-              the right edge of the last bin if a sequence.
+            * `bins`: if integer, number of equal-width bins; if sequence, bin edges that include the left edge of the
+              first bin and the right edge of the last bin.
             * `grid`: whether to show the grid lines.
             * `log`: set y-axis to log scale. If True, formatting defined in `y_ticker` is discarded.
 
-        Trace headers, whose values are measured in milliseconds (e.g. first break times) may be displayed over the
+        Trace headers, whose values are measured in milliseconds (e.g. first break times) may be displayed over a
         seismigram or wiggle plot if passed as `event_headers`. If `top_header` is passed, an auxiliary scatter plot of
         values of this header will be shown on top of the gather plot.
 
@@ -1078,8 +1078,8 @@ class Gather:
             * `step_labels`: place ticks with a given step between two adjacent ones in the units of the corresponding
               labels (e.g. place a tick every 200ms for `y` axis or every 300m offset for `x` axis). This option is
               valid only for "seismogram" and "wiggle" modes.
-        A short argument form allows defining both tickers as a single `str`, which will be treated as the value for
-        the `label` key. See :func:`~plot_utils.set_ticks` for more details on the ticker parameters.
+        A short argument form allows defining both tickers labels as a single `str`, which will be treated as the value
+        for the `label` key. See :func:`~plot_utils.set_ticks` for more details on the ticker parameters.
 
         Parameters
         ----------
@@ -1109,15 +1109,15 @@ class Gather:
         ax : matplotlib.axes.Axes, optional, defaults to None
             An axis of the figure to plot on. If not given, it will be created automatically.
         x_tick_src : str, optional
-            Source of the labels to be plotted on x axis. For "seismogram" and "wiggle" can be either "index" (default
+            Source of the tick labels to be plotted on x axis. For "seismogram" and "wiggle" can be either "index" (default
             if gather is not sorted) or any header; for "hist" it also defines the data source and can be either
             "amplitude" (default) or any header.
             Also serves as a default for axis label.
         y_tick_src : str, optional
-            Source of the data to be plotted on y axis. For "seismogram" and "wiggle" can be either "time" (default) or
+            Source of the tick labels to be plotted on y axis. For "seismogram" and "wiggle" can be either "time" (default) or
             "samples"; has no effect in "hist" mode. Also serves as a default for axis label.
         event_headers : str, array-like or dict, optional, defaults to None
-            Only for "seismogram" and "wiggle" modes.
+            Valid only for "seismogram" and "wiggle" modes.
             Headers, whose values will be displayed over the gather plot. Must be measured in milliseconds.
             If `dict`, allows controlling scatter plot options and handling outliers (header values falling out the `y`
             axis range). The following keys are supported:
@@ -1127,11 +1127,11 @@ class Gather:
                 * `discard`: do not display outliers,
                 * `none`: plot all the header values (default behavior).
             - Any additional arguments for `matplotlib.axes.Axes.scatter`.
-            If any dictionary value is array-like, each its element will be associated with the corresponding header.
+            If some dictionary value is array-like, each its element will be associated with the corresponding header.
             Otherwise, the single value will be used for all the scatter plots.
         top_header : str, optional, defaults to None
-            Only for "seismogram" and "wiggle" modes.
-            A header name to plot on top of the gather plot.
+            Valid only for "seismogram" and "wiggle" modes.
+            The name of a header whose values will be plotted on top of the gather plot.
         figsize : tuple, optional, defaults to (10, 7)
             Size of the figure to create if `ax` is not given. Measured in inches.
         save_to : str or dict, optional, defaults to None
@@ -1170,7 +1170,7 @@ class Gather:
         plotters_dict[mode](ax, title=title, x_ticker=x_ticker, y_ticker=y_ticker, **kwargs)
         return self
 
-    def _plot_histogram(self, ax, title=None, x_ticker=None, y_ticker=None, x_tick_src="amplitude", bins=None,
+    def _plot_histogram(self, ax, title, x_ticker, y_ticker, x_tick_src="amplitude", bins=None,
                         log=False, grid=True, **kwargs):
         """Plot histogram of the data scpecified by x_tick_src."""
         data = self.data if x_tick_src=="amplitude" else self[x_tick_src]
@@ -1184,7 +1184,7 @@ class Gather:
         ax.set_title(**{'label': None, **title})
 
     # pylint: disable=too-many-arguments
-    def _plot_seismogram(self, ax, title=None, x_ticker=None, y_ticker=None, x_tick_src=None, y_tick_src='time',
+    def _plot_seismogram(self, ax, title, x_ticker, y_ticker, x_tick_src=None, y_tick_src='time',
                          colorbar=False, qvmin=0.1, qvmax=0.9, event_headers=None, top_header=None, **kwargs):
         """Plot the gather as a 2d grayscale image of seismic traces."""
         # Make the axis divisible to further plot colorbar and header subplot
@@ -1201,7 +1201,7 @@ class Gather:
             ax.figure.colorbar(img, cax=cax, **colorbar)
         self._finalize_plot(ax, title, divider, event_headers, top_header, x_ticker, y_ticker, x_tick_src, y_tick_src)
 
-    def _plot_wiggle(self, ax, title='', x_ticker=None, y_ticker=None, x_tick_src=None, y_tick_src='time',
+    def _plot_wiggle(self, ax, title, x_ticker, y_ticker, x_tick_src=None, y_tick_src='time',
                      std=0.5, color="black", event_headers=None, top_header=None, **kwargs):
         """Plot the gather as an amplitude vs time plot for each trace."""
         # Make the axis divisible to further plot colorbar and header subplot
--- a/seismicpro/src/gather.py
+++ b/seismicpro/src/gather.py
@@ -1067,8 +1067,9 @@ class Gather:
         seismigram or wiggle plot if passed as `event_headers`. If `top_header` is passed, an auxiliary scatter plot of
         values of this header will be shown on top of the gather plot.
 
-        Ticks and their labels for both `x` and `y` axes can be controlled via `x_ticker` and `y_ticker` parameters
-        respectively. In the most general form, each of them is a `dict` with the following most commonly used keys:
+        While the source of label ticks for both `x` and `y` is defined by `x_tick_src` and `y_tick_src`, ticker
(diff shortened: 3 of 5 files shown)


nikita-klsh (commented):
As far as I understand, the motivation for such structure refactoring is the hist plot tickers. TBH the previous plot structure looks more clear to me. My suggestion how to unify the hist plot with other modes:

1. Hist plot becomes just another mode of `plotters_dict`
2.  Tickers inference  happens inside `plotters_dict[mode]`. Each mode knows how to infer it's own tickers(labels and title) (from headers or from `plt.hist`)
3. Method `_set_ticks` now only SETS tickers, and not infering them (inference happens on the prev. step)

seismicpro/src/utils/plot_utils.py
```diff
@@ -15,19 +15,17 @@ def save_figure(fig, fname, dpi=100, bbox_inches="tight", pad_inches=0.1, **kwar
     """Save the given figure. All `args` and `kwargs` are passed directly into `matplotlib.pyplot.savefig`."""
     fig.savefig(fname, dpi=dpi, bbox_inches=bbox_inches, pad_inches=pad_inches, **kwargs)
 
-
-def set_text_formatting(kwargs):
-    """Pop text formatting args from `kwargs` and set them as defaults for 'title', 'x_ticker' and 'y_ticker'."""
+def set_text_formatting(*args, **kwargs):
+    """Pop text formatting args from `kwargs` and set them as defaults for args transformed to dickts."""
```
alexandrkuvaev, line 19: dickts :)
Also, here you use word `args` twice in a single sentence with two different meanings

seismicpro/src/utils/plot_utils.py
```diff
@@ -15,19 +15,17 @@ def save_figure(fig, fname, dpi=100, bbox_inches="tight", pad_inches=0.1, **kwar
...
     FORMAT_ARGS = {'fontsize', 'size', 'fontfamily', 'family', 'fontweight', 'weight'}
-    TEXT_ARGS = {'title', 'x_ticker', 'y_ticker'}
 
     global_formatting = {arg: kwargs.pop(arg) for arg in FORMAT_ARGS if arg in kwargs}
-    text_args = {arg: {**global_formatting, **as_dict(kwargs.pop(arg), key="label")}
-                 for arg in TEXT_ARGS if arg in kwargs}
-    return {**kwargs, **text_args}
-
+    text_args = ({**global_formatting, **({} if arg is None else as_dict(arg, key="label"))}
+                  for arg in args)
```
alexandrkuvaev, line 24: I'd probably split it into two operations: first cast each `arg` to a dict and then merge it with defaults

seismicpro/src/utils/plot_utils.py
```diff
@@ -53,23 +51,32 @@ def set_ticks(ax, axis, axis_label, tick_labels, num=None, step_ticks=None, step
...
+        "Time": " (ms)",
+        "Offset": " (m)",
+    }
+    label = str.capitalize(label)
+    label += UNITS.get(label, "")
+
     ax_obj = getattr(ax, f"{axis}axis")
-    ax_obj.set_label_text(axis_label, **kwargs)
+    rotation_kwargs = _pop_rotation_kwargs(kwargs)
+    ax_obj.set_label_text(label, **kwargs)
```
alexandrkuvaev, line 64: Let's group all `ax_obj` processing lines together as it was before

seismicpro/src/utils/plot_utils.py
```diff
@@ -15,19 +15,17 @@ def save_figure(fig, fname, dpi=100, bbox_inches="tight", pad_inches=0.1, **kwar
...
-                 for arg in TEXT_ARGS if arg in kwargs}
-    return {**kwargs, **text_args}
-
+    text_args = ({**global_formatting, **({} if arg is None else as_dict(arg, key="label"))}
+                  for arg in args)
+    return text_args, kwargs
 
-def set_ticks(ax, axis, axis_label, tick_labels, num=None, step_ticks=None, step_labels=None, round_to=0, **kwargs):
+def set_ticks(ax, axis, label='', tick_labels=None, num=None, step_ticks=None,
+              step_labels=None, tick_range=None, round_to=0, **kwargs):
```
alexandrkuvaev, line 28: I'd place `tick_range` after `tick_labels`: all the following `kwargs` will generally be passed via `*_ticker`

seismicpro/src/utils/plot_utils.py
```diff
@@ -80,19 +87,37 @@ def _process_ticks(labels, num, step_ticks, step_labels, round_to):
...
+        # issue.
+        if len(labels) == 1:
+            labels = np.repeat(labels, 2)
+        # matplotlib does not update data interval when new artist is redrawn on the existing axes in interactive mode,
+        # which leads to incorrect tick position to label interpolation. To overcome this, call `ax.clear()` before
+        # drawing a new artist.
+        tick_interpolator = interp1d(np.linspace(*tick_range, len(labels)), labels,
+                                     kind="nearest", bounds_error=False)
+
+        def formatter(label_ix, *args):
```
alexandrkuvaev, line 111: Unify `formatter` signature and move it outside if-else block next to `round_formatter`

seismicpro/src/gather.py
```diff
@@ -1195,10 +1203,31 @@ def _plot_wiggle(self, ax, std=0.5, color="black"):
         y_coords = np.arange(self.n_samples)
         traces = std * (self.data - self.data.mean(axis=1, keepdims=True)) / (np.std(self.data) + 1e-10)
         for i, (trace, col) in enumerate(zip(traces, color)):
-            ax.plot(i + trace, y_coords, color=col)
-            ax.fill_betweenx(y_coords, i, i + trace, where=(trace > 0), color=col)
+            ax.plot(i + trace, y_coords, color=col, **kwargs)
+            ax.fill_betweenx(y_coords, i, i + trace, where=(trace > 0), color=col, **kwargs)
```
alexandrkuvaev, line 1207: Have you checked that identical `kwargs` can be passed to both `plot` and `fill_between`?

alexandrkuvaev (changes requested):

seismicpro/src/gather.py
```diff
@@ -1143,37 +1142,41 @@ def plot(self, mode="seismogram", event_headers=None, top_header=None, title=Non
...
-        if x_ticker is None:
-            x_ticker = self.sort_by if self.sort_by is not None else "index"
-        self._set_ticks(ax, axis="x", ticker=x_ticker)
-        self._set_ticks(ax, axis="y", ticker=y_ticker)
+        ax = plotters_dict[mode](ax, x_ticker=x_ticker, y_ticker=y_ticker, **kwargs)
+        ax.set_title(**{'label': None, **title})
         return self
 
-    def _plot_seismogram(self, ax, divider, colorbar=False, qvmin=0.1, qvmax=0.9, **kwargs):
+    def _plot_histogram(self, ax, bins=50, x_tick_src="amplitude", log=False, x_ticker=None, y_ticker=None, grid=False,
```
anton-br, line 1160: don't like `src` in `x_tick_src` since it is not a batch and we don't have any src here. But cannot provide any other namings for now.

seismicpro/src/batch.py
```diff
@@ -487,12 +487,12 @@ def plot(self, src, src_kwargs=None, max_width=20, title="{src}: {index}", save_
 
                 # Format subplot title
                 if title_template is not None:
-                    title = as_dict(title_template, key='label')
-                    label = title.pop("label")
+                    src_title = as_dict(title_template, key='label')
```
anton-br, line 490: why don't you like just `title`? We don't have other titles here

seismicpro/src/gather.py
dpodvyaznikov: Well, `*_tick_src` identifies the source of the tick data for the axis, which makes sense.

seismicpro/src/batch.py
dpodvyaznikov: Actually, we have - the one that is passed in signature and used to build `kwargs` for each `src`. It was redefined here which led to incorrect titles for plots of the second and following `src`'s.

seismicpro/src/decorators.py
```diff
@@ -57,7 +57,6 @@ def plotter(figsize, args_to_unpack=None):
     def decorator(method):
         @wraps(method)
         def plot(*args, **kwargs):
-            kwargs = set_text_formatting(kwargs)
```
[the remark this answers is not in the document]
dpodvyaznikov: This would allow for more flexibility in plotters, since now `set_text_formatting` can be applied not only to `title` and `*_ticker`, but to any text-related parameters of the plot, e.g., legends or any other text caption on the axes.

seismicpro/src/utils/plot_utils.py
dpodvyaznikov: Like this? I don't fancy it, but would switch to this one If others agree

```suggestion
    dict_args = ({} if arg is None else as_dict(arg, key="label") for arg in args)
    text_args = ({**global_formatting, **arg} for arg in dict_args)
```

seismicpro/src/gather.py
dpodvyaznikov: This change seemed to make sense since the most common parameters, like `'label'`, `'linewidth'`, `'linestyle'`, `'color'` are identical.

seismicpro/src/utils/plot_utils.py
dpodvyaznikov: See in next commit

seismicpro/src/utils/plot_utils.py
dpodvyaznikov: See in next commit

seismicpro/src/utils/plot_utils.py
dpodvyaznikov: See in next commit

seismicpro/src/gather.py
```diff
@@ -1195,10 +1203,31 @@ def _plot_wiggle(self, ax, std=0.5, color="black"):
...
+        if event_headers is not None:
+            self._plot_headers(ax, event_headers)
+
+        # Add a top subplot for given header if needed and set plot title
+        top_ax = ax
+        if top_header is not None:
+            top_ax = self._plot_top_subplot(ax=ax, divider=divider, header_values=self[top_header].ravel())
+
+        # Set axis ticks
+        x_tick_src = (self.sort_by if self.sort_by is not None else "index") if x_tick_src is None else x_tick_src
```
a-arefina, line 1225:
```suggestion
        x_tick_src = x_tick_src or self.sort_by or "index" 
```

seismicpro/src/gather.py
```diff
@@ -1195,10 +1203,31 @@ def _plot_wiggle(self, ax, std=0.5, color="black"):
...
+
+    def _finalize_plot(self, ax, divider, event_headers, top_header, x_ticker, y_ticker, x_tick_src, y_tick_src):
+        # Add headers scatter plot if needed
+        if event_headers is not None:
+            self._plot_headers(ax, event_headers)
+
+        # Add a top subplot for given header if needed and set plot title
+        top_ax = ax
+        if top_header is not None:
+            top_ax = self._plot_top_subplot(ax=ax, divider=divider, header_values=self[top_header].ravel())
```
a-arefina, line 1222: you're returning top_ax to place a plot title there? 
Maybe move `ax.set_title(**{'label': None, **title})` from `plot` to `_finalize_plot`?

Or, at least add a comment

seismicpro/src/semblance.py
```diff
@@ -184,10 +187,8 @@ def plot(semblance, title=None, x_label=None, x_ticklabels=None,  # pylint: disa
         if grid:
             ax.grid(c='k')
 
-        x_ticker = kwargs if x_ticker is None else {**kwargs, **x_ticker}
-        y_ticker = kwargs if y_ticker is None else {**kwargs, **y_ticker}
         set_ticks(ax, "x", x_label, x_ticklabels, **x_ticker)
-        set_ticks(ax, "y", "Time (ms)", y_ticklabels, **y_ticker)
+        set_ticks(ax, "y", "Time", y_ticklabels, **y_ticker)
```
a-arefina, line 191: why did you remove '(ms)' ?

nikita-klsh (commented):
How fully does hist mode support our ticker parametrization?

1. In this case `bins=10` and `step_ticks=1`. We expect that each of 10 bin would have it's own tick?

image
2. y locators has no effect when `log=True`
image
image
3. y locators not working correctly when `log=False`
image
image
4. 
image

seismicpro/src/gather.py
dpodvyaznikov: Yes
I don't think that title setting should be in `finalize_plot`. Moreover - finalize_plot is not called in `_plot_histogram`, so we'd also need to call it there.

seismicpro/src/gather.py
```diff
@@ -1195,10 +1203,31 @@ def _plot_wiggle(self, ax, std=0.5, color="black"):
...
-            ax.fill_betweenx(y_coords, i, i + trace, where=(trace > 0), color=col)
+            ax.plot(i + trace, y_coords, color=col, **kwargs)
+            ax.fill_betweenx(y_coords, i, i + trace, where=(trace > 0), color=col, **kwargs)
         ax.invert_yaxis()
 
+        # Wiggle plot requires custom data interval for correct tick setting
+        x_ticker.update({"tick_range":(0, self.n_traces-1)})
+        return self._finalize_plot(ax, divider, event_headers, top_header, x_ticker, y_ticker, x_tick_src, y_tick_src)
+
+    def _finalize_plot(self, ax, divider, event_headers, top_header, x_ticker, y_ticker, x_tick_src, y_tick_src):
```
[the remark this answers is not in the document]
dpodvyaznikov: I don't mind renaming this function, but don't want to move tick setting. 
What do others think? @alexandrkuvaev @a-arefina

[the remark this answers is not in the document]
dpodvyaznikov: I don't mind renaming this function, but don't want to move tick setting. 
What do others think? @alexandrkuvaev @a-arefina

seismicpro/src/semblance.py
dpodvyaznikov: It is going to be added in `set_tick`

seismicpro/src/utils/plot_utils.py
```diff
@@ -80,19 +87,37 @@ def _process_ticks(labels, num, step_ticks, step_labels, round_to):
...
+    def round_formatter(label_value, *args):
         _ = args
-        if (label_ix < 0) or (label_ix > len(labels) - 1):
-            return None
-
-        label_value = labels[np.round(label_ix).astype(np.int32)]
         if round_to is not None:
             label_value = np.round(label_value, round_to)
             label_value = label_value.astype(np.int32) if round_to == 0 else label_value
         return label_value
```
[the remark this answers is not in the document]
dpodvyaznikov: Current implementation for `label_value=1.1` with `round_to=3` produces '1.1', while suggested option would produce '1.100', which may be excessive.
Moreover, in current implementation one can set `round_to=None` and get the full representation for each label regardless of number of decimals.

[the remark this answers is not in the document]
nikita-klsh: For me getting 1.100 is what I expect passing `round_to=3`, if I understand `round_to` functionality correctly.
Here is how `round_to` works atm, it looks confusing to me
image

Refering to the [documentation](https://github.com/matplotlib/matplotlib/blob/710fce3df95e22701bd68bf6af2c8adbc9d67a79/lib/matplotlib/ticker.py#L312-L337), FuncFormatter is supposed to be instantiated by the function, which produces a string. We, however, return numpy scalar, which is somehow formatted latter(in different ways, as can be seen on the pic)

Btw my suggestion behaves exactly like the current in case of `None`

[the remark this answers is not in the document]
dpodvyaznikov: Yep, my bad: in cases of `0` or `None` it works similarly, I'll use your version

seismicpro/src/gather.py
dpodvyaznikov: Done

seismicpro/src/gather.py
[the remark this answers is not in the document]
a-arefina: I also wanted to propose renaming but got stuck with a new name :(

I like the idea of separate functions that set ticks, draw optional axes, and set titles where needed. As I commented above, I feel rather uneasy that this function returns an axis, that is further returned to `plot` with a sole aim of setting a title at the correct place.

seismicpro/src/gather.py
dpodvyaznikov: Done

dpodvyaznikov:
Replies for the Nikita's comment:
1. In this case `bins=10` and `step_ticks=1`. We expect that each of 10 bin would have it's own tick?
— step_ticks work in the units of the x-axis for hist plot. if the data range is 0-1 and you specify `step_ticks=0.1`, you'll get 10 ticks.
2. y locators has no effect when `log=True`
— this is intended and mentioned in docstrings.
3. y locators not working correctly when `log=False`
— This behavior is similar with  `wiggle` and `seismogram` modes. The problem is caused by `num` argument and LinearLocator - it usually sets from `num` to `num-2` ticks, depending on the data intervals and some edge cases.
4. step_labels have no effect
— in `hist` mode `tick_labels` are None, so that was intended. For clarity, now if `labels=None` and `step_labels` is specified, an error is raised.

seismicpro/src/gather.py
```diff
@@ -1183,9 +1199,14 @@ def _plot_seismogram(self, ax, divider, colorbar=False, qvmin=0.1, qvmax=0.9, **
             colorbar = {} if colorbar is True else colorbar
             cax = divider.append_axes("right", size="5%", pad=0.05)
             ax.figure.colorbar(img, cax=cax, **colorbar)
+        self._finalize_plot(ax, title, divider, event_headers, top_header, x_ticker, y_ticker, x_tick_src, y_tick_src)
 
-    def _plot_wiggle(self, ax, std=0.5, color="black"):
+    def _plot_wiggle(self, ax, title='', x_ticker=None, y_ticker=None, x_tick_src=None, y_tick_src='time',
```
nikita-klsh, line 1204: Can we unify default value `title=None` like in other modes?

nikita-klsh, line 1204: Can we actually make `ax, title, x_ticker and y_ticker` required positional args for all the modes. Now we just repeating defaults from `self.plot`

seismicpro/src/gather.py
```diff
@@ -1119,6 +1108,30 @@ def plot(self, mode="seismogram", event_headers=None, top_header=None, title=Non
             formatting and layout, see :func:`~plot_utils.set_ticks` for more details.
         ax : matplotlib.axes.Axes, optional, defaults to None
             An axis of the figure to plot on. If not given, it will be created automatically.
+        x_tick_src : str, optional
+            Source of the labels to be plotted on x axis. For "seismogram" and "wiggle" can be either "index" (default
```
nikita-klsh, line 1112:
```suggestion
            Source of the tick labels to be plotted on x axis. For "seismogram" and "wiggle" can be either "index" (default
```

seismicpro/src/gather.py
```diff
@@ -1119,6 +1108,30 @@ def plot(self, mode="seismogram", event_headers=None, top_header=None, title=Non
...
+            if gather is not sorted) or any header; for "hist" it also defines the data source and can be either
+            "amplitude" (default) or any header.
+            Also serves as a default for axis label.
+        y_tick_src : str, optional
+            Source of the data to be plotted on y axis. For "seismogram" and "wiggle" can be either "time" (default) or
```
nikita-klsh, line 1117:
```suggestion
            Source of the tick labels to be plotted on y axis. For "seismogram" and "wiggle" can be either "time" (default) or
```

nikita-klsh (approved):
Please modify docs for `x_ticker` and `y_ticker`, now they do not control source for tick labels.

seismicpro/src/utils/plot_utils.py
```diff
@@ -36,28 +33,46 @@ def set_ticks(ax, axis, axis_label, tick_labels, num=None, step_ticks=None, step
...
-    rotation_kwargs = _pop_rotation_kwargs(kwargs)
+    # Format axis label
+    UNITS = {  # pylint: disable=invalid-name
+        "Time": " (ms)",
+        "Offset": " (m)",
+    }
+    label = label[0].upper() + label[1:]
+    label += UNITS.get(label, "")
+
     ax_obj = getattr(ax, f"{axis}axis")
```
alexandrkuvaev, line 71: Move to `ax_obj` processing lines

alexandrkuvaev (approved):

seismicpro/src/utils/plot_utils.py
```diff
@@ -15,19 +16,15 @@ def save_figure(fig, fname, dpi=100, bbox_inches="tight", pad_inches=0.1, **kwar
     """Save the given figure. All `args` and `kwargs` are passed directly into `matplotlib.pyplot.savefig`."""
     fig.savefig(fname, dpi=dpi, bbox_inches=bbox_inches, pad_inches=pad_inches, **kwargs)
 
-
-def set_text_formatting(kwargs):
-    """Pop text formatting args from `kwargs` and set them as defaults for 'title', 'x_ticker' and 'y_ticker'."""
+def set_text_formatting(*args, **kwargs):
+    """Pop text formatting parameters from `kwargs` and set them as defaults for each of `args` tranformed to dict."""
     FORMAT_ARGS = {'fontsize', 'size', 'fontfamily', 'family', 'fontweight', 'weight'}
```
alexandrkuvaev, line 21: Let's remove all the synonyms and keep only parameters, starting with `font`

alexandrkuvaev (approved):

PR #84, opened 2022-02-15
fix speed, add test params

a-arefina:
Dead traces detection works slowly compared to `collect_stats` due to slow work of `np.isclose` for scalars (https://github.com/numpy/numpy/issues/16160) 

I've replaced np.isclose with math.isclose in `mark_dead_traces` and added various survey indices in dead traces related tests

 here (https://github.com/numpy/numpy/issues/10161) some differences in np.isclose and math.isclose are discussed

### DIFF
--- a/seismicpro/src/survey.py
+++ b/seismicpro/src/survey.py
@@ -5,6 +5,7 @@ import os
 import warnings
 from copy import copy, deepcopy
 from textwrap import dedent
+import math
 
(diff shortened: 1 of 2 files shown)


dpodvyaznikov (approved):
I approve this PR, but, probably, we'll need to change the overall logic of dead trace detection when developing new QC  metrics

PR #93, opened 2022-04-05
Update trace headers loading logic

alexanderkuvaev:
Speed up survey headers loading by switching to a tracewise iteration over the SEG-Y file. This approach:
* improves cache utilization for large files thus speeding up their loading,
* allows for adding a meaningful progress bar.

Caveat: small surveys (~100k traces) are now loaded a little bit slower (~700ms vs ~100ms). Doesn't make a big difference, but worth keeping in mind.


seismicpro/src/survey/utils.py
```diff
@@ -1,11 +1,107 @@
...
+
+
+def read_headers_chunk(path, chunk_offset, chunk_size, trace_stride, headers_format, endian):
+    """Read `chunk_size` trace headers starting from `chunk_offset` byte in the SEG-Y file."""
+    with open(path, "rb") as f:
+        with mmap.mmap(f.fileno(), length=0, access=mmap.ACCESS_READ) as mm:
+            headers = np.ndarray(buffer=mm, dtype=np.dtype(f"V{TRACE_HEADER_SIZE}"), offset=chunk_offset,
+                                 shape=chunk_size, strides=trace_stride).tolist()
+    # Construct a format string for the whole chunk
+    chunk_format_str = {"big": ">", "msb": ">", "little": "<", "lsb": "<"}[endian] + headers_format * chunk_size
```
SergeyTsimfer, line 65: Maybe, move some of the constant mappings to constants of this file?
Also, maybe move all three functions to a separate `headers.py`?

seismicpro/src/survey/utils.py
```diff
@@ -1,11 +1,107 @@
...
+    def shutdown(self, *args, **kwargs):
+        """Signal the executor to finish all scheduled tasks and free its resources."""
+        _ = args, kwargs
+        for future, fn in self.task_queue:
+            future.set_result(fn())
+        self.task_queue = None
+
+
+def define_unpacking_format(headers_to_load):
+    """Return a format string to unpack `headers_to_load` trace headers from a headers byte sequence."""
```
SergeyTsimfer, line 43: Give an example in the doc

SergeyTsimfer (approved):

seismicpro/src/survey/utils.py
alexanderkuvaev: Clarified the structure. Example is not provided since the string is usually ~200 chars long

anton-br (approved):

seismicpro/src/utils/general_utils.py
```diff
@@ -182,3 +185,26 @@ def __getattr__(self, name):
     def __call__(self, *args, **kwargs):
         _ = args, kwargs
         raise ImportError(f"No module named {self._module_name}")
+
+
+class ForPoolExecutor(Executor):
+    """A sequential executor of tasks in a for loop. Inherits `Executor` interface thus can serve as a seamless
```
dpodvyaznikov, line 191:
```suggestion
    """A sequential executor of tasks in a for loop. Inherits `Executor` interface thus can serve as a drop-in
```

dpodvyaznikov (approved):

PR #97, opened 2022-05-25
Speed up trace loading

alexanderkuvaev:
Major improvements:
* Seismic trace loading:
  * Now uses `numpy.memmap` by default instead of `segyio` interface. Loading now ~3 times faster using `get_gather` and ~2 times faster in pipelines.
  * Added compatibility mode which forces old loading algorithm. Can be turned on by passing `use_segyio_trace_loader=True` during survey instantiation.
  * Fixed a bug in current trace loading algoritm: the buffer to store the trace was always created with `float32` dtype. This resulted in wrong loaded amplitudes if the source SEG-Y file stored them in any other format (which it actually can!).
* `collect_stats`:
  * Fixed all potential overflows: previously `Survey.std` was calculated using sum of squares of traces amplitudes which is erroneous for large surveys or surveys with large absolute amplitudes.
  * Refactored to load and process traces in chunks. Combined with improved trace loading sped up stats collection for a survey of 150kk traces from 15 hours to 30 minutes. Small surveys are processed ~3 times faster.

Minor fixes:
* `UnassignedInt1` and `UnassignedInt2` trace headers are now forbidden to be loaded since they are treated differently from all other headers by `segyio`. For example, `segyio` ignores SEG-Y file endiannes during their loading and hides them during headers access, which made them equal 0 after each `aggregate_segys`.
* Rename `copy_header` argument in `Gather.dump` to `retain_parent_segy_headers` and set its default value to `True`. Old name was inconsistent with what it really did. Moreover, it conflicted with `copy_headers` argument in almost all our loaders/constructors, which actually copies headers.
* `Gather.get_central_cdp` was renamed to `get_central_gather` to be more precise.
* `Gather.stack` now preserves trace headers of the first trace of the gather being stacked, which allows dumping not only `INLINE_3D` and `CROSSLINE_3D`, but also `CDP_X` and `CDP_Y`.


nikita-klsh:
Can you please update the stacking pipeline on the main page of SeismicPro with new interface: `get_central_gather `and `dump`. Overall the pipeline is working fine

alexanderkuvaev:
> Can you please update the stacking pipeline on the main page of SeismicPro with new interface: `get_central_gather `and `dump`. Overall the pipeline is working fine

Fixed, ty

dpodvyaznikov (approved):

PR #107, opened 2022-10-05
Calculate number of refractors for the RefractorVelocityField.

bestetc:
Now:
If a RefractorVelocityField is created using the `from_survey` classmethod with no initial values, bounds, or number of refractors, the number of refractors and initial values will be calculated from the offsets and times of the first breaks stored in Survey.


seismicpro/src/refractor_velocity/utils.py
```diff
@@ -86,3 +87,68 @@ def load_refractor_velocity_params(path, encoding="UTF-8"):
...
+        else:
+            break
+    if plot_last and rv is not None:  # debug feature
+        rv.plot(title=name)
+    return rv
+
+def calc_optimal_velocity(survey, min_offsets_diff=300, min_velocity_diff=300, fb_col=HDR_FIRST_BREAK, weathering=True,
+                     name=None, plot_last=False):
+    """Calculate one velocity model describe passed survey."""
+    if len(survey.indices) < 1:
```
dpodvyaznikov, line 135: `survey` has `__len__`:
```suggestion
    if len(survey) < 1:
```

seismicpro/src/refractor_velocity/utils.py
```diff
@@ -86,3 +87,68 @@ def load_refractor_velocity_params(path, encoding="UTF-8"):
...
+        raise ValueError("Object is empty")
```
dpodvyaznikov, line 136:
```suggestion
        raise ValueError("survey is empty")
```

seismicpro/src/refractor_velocity/utils.py
```diff
@@ -86,3 +87,68 @@ def load_refractor_velocity_params(path, encoding="UTF-8"):
...
+        n_points, _ = np.histogram(rv_last.offsets, bins=rv_last.piecewise_offsets)
+        if (n_points > 1).all() and (rv is None or rv_last.fit_result.fun < rv.fit_result.fun):
+            rv = rv_last
+        else:
+            break
+    if plot_last and rv is not None:  # debug feature
+        rv.plot(title=name)
+    return rv
+
+def calc_optimal_velocity(survey, min_offsets_diff=300, min_velocity_diff=300, fb_col=HDR_FIRST_BREAK, weathering=True,
```
dpodvyaznikov, line 132: In the rest of the codebase we use `first_breaks_col`

seismicpro/src/refractor_velocity/utils.py
```diff
@@ -86,3 +87,68 @@ def load_refractor_velocity_params(path, encoding="UTF-8"):
...
+                     name=None, plot_last=False):
+    """Calculate one velocity model describe passed survey."""
+    if len(survey.indices) < 1:
+        raise ValueError("Object is empty")
+    offsets = survey.headers['offset'].ravel()
```
dpodvyaznikov, line 137: Wouldn't it be shorter and faster to use `survey['offset']`?

seismicpro/src/refractor_velocity/utils.py
bestetc: `survey.n_gathers` prefer here.

seismicpro/src/refractor_velocity/utils.py
```diff
@@ -86,3 +87,68 @@ def load_refractor_velocity_params(path, encoding="UTF-8"):
         params['coords'] = Coordinates(names=tuple(row[:2]), coords=tuple(row[2:4].astype(int)))
         params_list.append(params)
     return params_list
+
+# calculate number of refractors
+
+def binarization_offsets(offsets, times, step=20):
+    """Binarize offsets-times points."""
+    bins = np.arange(0, offsets.max() + step, step=step)
+    mean_offsets = np.arange(bins.shape[0]) * step + step / 2
```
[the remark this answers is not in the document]
bestetc: fully rework binarize

seismicpro/src/refractor_velocity/refractor_velocity.py
```diff
@@ -398,8 +398,11 @@ def _validate_params(cls, params, max_offset=None, min_velocity_step=0, min_refr
         if np.any(np.diff(param_values[1:n_refractors], prepend=0, append=max_offset) < min_refractor_size):
             raise ValueError(f"Offset range covered by refractors must be no less than {min_refractor_size} meters")
 
-        if np.any(np.diff(param_values[n_refractors:]) < min_velocity_step):
-            raise ValueError(f"Refractor velocities must increase by no less than {min_velocity_step} m/s")
+        # if np.any(np.diff(param_values[n_refractors:]) < min_velocity_step):
+        if np.any((np.diff(param_values[n_refractors:]) < min_velocity_step) *
```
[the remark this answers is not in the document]
bestetc: Similar. But scipy  precision sometimes it not.
I'd test it on master branch with:
```
offsets = np.linspace(0, 1500, num=100)
times = offsets * .67 + 100

for i in range(100):
    try:
        rv = RefractorVelocity.from_first_breaks(offsets, times, n_refractors=3, min_velocity_step=300)
    except ValueError as er:
        print(f"{i} " + str(er))
```

seismicpro/src/refractor_velocity/utils.py
```diff
@@ -73,3 +74,117 @@ def load_refractor_velocities(path, encoding="UTF-8"):
...
+
+    Returns
+    -------
+    rv : RefractorVelocity or None
+        A near-surface velocity model with optimal number of refractors. Returns None if no velocity model is possible
+        for the given parameters.
+    """
+    #pylint: disable-next=import-outside-toplevel
+    from .refractor_velocity import RefractorVelocity  # avoid circulat import
+    rv = None
```
dpodvyaznikov, line 124: Unless offsets and times contain less than 2 points, you should be able to make a RV with 1 refractor. I'd suggest fitting RV with `n_refractors = 1` here, outside of the loop and use it as a default.
This would also make stopping criteria below much simpler. Instead of 
```
        if not ((n_points > 1).all() and (rv is None or rv_last.fit_result.fun < rv.fit_result.fun)):
            break
```
you'd have
```
if (n_points < 1).any() or (rv.fit_result.fun - rv_last.fit_result.fun < some_tol )):
            break
```

dpodvyaznikov: You need to use `some_tol` to compare losses since there is a stopping criteria and beyond it all the losses are basicly the same

seismicpro/src/refractor_velocity/utils.py
```diff
@@ -73,3 +74,117 @@ def load_refractor_velocities(path, encoding="UTF-8"):
...
+    ------
+    ValueError
+        If survey does not contain any indices.
+    """
+    if survey.n_gathers < 1:  # need if the func calls separately from `RefractorVelocityField.from_survey`
+        raise ValueError("Survey is empty.")
+    offsets, times = reduce_mean_df(survey.headers[['offset', first_breaks_col]])
+    rv = calc_optimal_velocity(offsets, times, min_refractor_size=min_refractor_size,
+                               min_velocity_step=min_velocity_step, debug=debug)
+    if find_weathering:  # try to find the weathering layer
```
dpodvyaznikov, line 182: Talk to @alexanderkuvaev, maybe it would be better to remove weathering calculation at all

dpodvyaznikov (changes requested):

seismicpro/src/refractor_velocity/refractor_velocity.py
```diff
@@ -406,7 +406,8 @@ def _validate_params(cls, params, max_offset=None, min_velocity_step=0, min_refr
         if np.any(np.diff(param_values[1:n_refractors], prepend=0, append=max_offset) < min_refractor_size):
             raise ValueError(f"Offset range covered by refractors must be no less than {min_refractor_size} meters")
 
-        if np.any(np.diff(param_values[n_refractors:]) < min_velocity_step):
+        if np.any((np.diff(param_values[n_refractors:]) < min_velocity_step) *
```
dpodvyaznikov, line 409: It is an overly complex and unintuitive solution to the problem.

The logic of this check is simple: make sure that the difference is no less that some value. After some digging, you see that sometimes this check fails, most likely due to float numer representation precision. 
Since strict `<` check does not work, so you add second `np.isclose` check - basicly a comparison with some tolerance.
Why not use this tolerance right away in `<` comparison?
```if np.any((np.diff(param_values[n_refractors:]) < (1-rtol) * min_velocity_step)):```
or
```if np.any((np.diff(param_values[n_refractors:]) < (min_velocity_step - atol))):```

Basicly the same check that you wrote, but simpler. Moreover, `np.isclose` docs state that comparing values much less than 1 with default `atol` parameter is dicsouraged.

Update: this constraint can be enforced in `refractor_velocity.utils.postprocess_params` function, similar to `RefractorVelocity.enforce_step_constraints`. Consult with @alexanderkuvaev

**Review by dpodvyaznikov — CHANGES_REQUESTED** [review]

File: seismicpro/src/survey/survey.py [hunk 10]
```diff
@@ -1247,3 +1248,27 @@ def construct_attribute_map(self, attribute, by, drop_duplicates=False, agg=None
 
         metric = PartialMetric(SurveyAttribute, survey=self, name=attribute, **kwargs)
         return metric.map_class(map_data.iloc[:, :2], map_data.iloc[:, 2], metric=metric, agg=agg, bin_size=bin_size)
+
+    def plot_mean_velocity(self, loss="L1", huber_coef=20, min_velocity_step=300, min_refractor_size=300,
```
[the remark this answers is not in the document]
bestetc: @nikita-klsh ask me to add this method.

seismicpro/src/refractor_velocity/utils.py
```diff
@@ -73,3 +74,117 @@ def load_refractor_velocities(path, encoding="UTF-8"):
...
+        A near-surface velocity model with optimal number of refractors. Returns None if no velocity model is possible
+        for the given parameters.
+    """
+    #pylint: disable-next=import-outside-toplevel
+    from .refractor_velocity import RefractorVelocity  # avoid circulat import
+    rv = None
+    for refractor in range(min_refractors, max_refractors + 1):
+        min_refractor_size_vec = np.full(refractor, min_refractor_size)
+        if find_weathering:
+            min_refractor_size_vec[0] = 1
```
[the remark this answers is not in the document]
bestetc: Yes, when I call `calc_optimal_velocity` directly.
Usually `calc_optimal_velocity` call with `fing_weathering` flag by `calc_mean_velocity` with `start_refractor` at least 2.

seismicpro/src/refractor_velocity/refractor_velocity_field.py
```diff
@@ -205,6 +206,8 @@ def from_survey(cls, survey, is_geographic=None, auto_create_interpolator=True,
         """
         if survey.n_gathers < 1:
             raise ValueError("Survey is empty.")
+        if all(param is None for param in (init, bounds, n_refractors)):
+            init = calc_mean_velocity(survey, first_breaks_col=HDR_FIRST_BREAK, find_weathering=True).params
```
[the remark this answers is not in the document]
bestetc: by design, I don't want to pass any parameters from `from_survey` to `calc_mean_velocity`. If you want to make another input you could call `calc_mean_velocity` directly.

seismicpro/src/survey/survey.py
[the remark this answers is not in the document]
nikita-klsh: Make `survey.estimate_mean_velocity` method which returns `RefractorVelocity`.

seismicpro/src/refractor_velocity/utils.py
```diff
@@ -73,3 +74,117 @@ def load_refractor_velocities(path, encoding="UTF-8"):
...
+    ValueError
+        If survey does not contain any indices.
+    """
+    if survey.n_gathers < 1:  # need if the func calls separately from `RefractorVelocityField.from_survey`
+        raise ValueError("Survey is empty.")
+    offsets, times = reduce_mean_df(survey.headers[['offset', first_breaks_col]])
+    rv = calc_optimal_velocity(offsets, times, min_refractor_size=min_refractor_size,
+                               min_velocity_step=min_velocity_step, debug=debug)
+    if find_weathering:  # try to find the weathering layer
+        init = {'x1': 150, 'v1': rv.v1 / 2}
```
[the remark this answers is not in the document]
bestetc: `calc_optimal_velocity` always retrun rv if `start_refractor` is 1. It could be if I try to find weathering layer with `start_refractor` more than 1.

seismicpro/src/refractor_velocity/refractor_velocity.py
bestetc: I used `isclose` to made two check together. But your decision are simpler and could be combine:
`if np.any((np.diff(param_values[n_refractors:]) < (1-rtol) * min_velocity_step) - atol):`

I do not wait here the velocities much less than 1.

seismicpro/src/refractor_velocity/refractor_velocity_field.py
```diff
@@ -207,6 +208,8 @@ def from_survey(cls, survey, is_geographic=None, init=None, bounds=None, n_refra
         """
         if survey.n_gathers < 1:
             raise ValueError("Survey is empty.")
+        if all(param is None for param in (init, bounds, n_refractors)):
```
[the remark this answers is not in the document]
bestetc: `param` says me a little bit more than `_`.

seismicpro/src/survey/survey.py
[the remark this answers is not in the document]
dpodvyaznikov: This should not be in the `Survey`, let's make `RafractorVelocity.from_survey` method. 
This way it would be easy to use `calc_mean_velocity` without `RafractorVelocityField`, and plotting this "average" velocity would be simply:
`RafractorVelocity.from_survey(some_survey).plot()`

seismicpro/refractor_velocity/utils.py
```diff
@@ -91,3 +92,121 @@ def dump_refractor_velocities(refractor_velocities, path, encoding="UTF-8"):
...
+            min_refractor_size_vec[0] = 1
+        max_offset = max(offsets.max(), min_refractor_size * refractor)
+        rv_last = RefractorVelocity.from_first_breaks(offsets, times, init, bounds, refractor, max_offset,
+                                                      min_velocity_step, min_refractor_size_vec, loss, huber_coef)
+        n_points, _ = np.histogram(rv_last.offsets, bins=rv_last.piecewise_offsets)
+        # TODO: remove debug
+        if debug:
+            rv_last.plot(title=(f'loss: {rv_last.fit_result.fun:.6f}\nfind_weathering={find_weathering}'
+                                f'\nn_points: {n_points}'))
+        if not ((n_points > 1).all() and (rv is None or rv_last.fit_result.fun < rv.fit_result.fun)):
```
anton-br, line 158: If you don't want to add toleranceas @dpodvyaznikov suggested, at least simplify first part to 
```suggestion
        if (n_points < 1).any() and (rv is None or rv_last.fit_result.fun < rv.fit_result.fun)):
```

seismicpro/refractor_velocity/utils.py
```diff
@@ -91,3 +92,121 @@ def dump_refractor_velocities(refractor_velocities, path, encoding="UTF-8"):
         data[i] = [*rv.coords.names] + [*rv.coords.coords] + list(rv.params.values())
     df = pd.DataFrame(data, columns=columns).convert_dtypes()
     df.to_string(buf=path, float_format=lambda x: f"{x:.2f}", index=False, encoding=encoding)
+
+
+def reduce_survey_headers(survey, x="offset", y=HDR_FIRST_BREAK, reduce_step=20):
```
anton-br, line 97: Consider using this function only when x and y are stings. So delete to_list and ravel inside.

seismicpro/refractor_velocity/utils.py
```diff
@@ -91,3 +92,122 @@ def dump_refractor_velocities(refractor_velocities, path, encoding="UTF-8"):
...
+    if survey.n_gathers < 1:
+        raise ValueError("Survey is empty.")
+    offsets, times = reduce_survey_headers(survey, x="offset", y=first_breaks_col, reduce_step=reduce_step)
+    if offsets.shape[0] < 2:
+        raise ValueError("Offset contains less than two points after reducing. Decrease the value of `reduce_step`.")
+    rv = calc_optimal_velocity(offsets, times, min_velocity_step=min_velocity_step,
+                               min_refractor_size=min_refractor_size, loss=loss, huber_coef=huber_coef, debug=debug)
+    if find_weathering:
+        init = {'x1': 150, 'v1': rv.v1 / 2}
+        bounds = {'x1': [1, 300], 'v1': [1, rv.v1]}
```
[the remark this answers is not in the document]
bestetc: I do not wait here the velocities less than 1 meters per second.

seismicpro/refractor_velocity/refractor_velocity.py
```diff
@@ -401,7 +416,7 @@ def _validate_params(cls, params, max_offset=None, min_velocity_step=0, min_refr
         if np.any(np.diff(param_values[1:n_refractors], prepend=0, append=max_offset) < min_refractor_size):
             raise ValueError(f"Offset range covered by refractors must be no less than {min_refractor_size} meters")
 
-        if np.any(np.diff(param_values[n_refractors:]) < min_velocity_step):
+        if np.any((np.diff(param_values[n_refractors:]) < (1 - 1e-5) * min_velocity_step - 1e-8)):
```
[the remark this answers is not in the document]
dpodvyaznikov: This suggestion won't work for 2 reasons:
1) You don't want to compare velocities - `param_values[n_refractors:]` and `min_velocity_step`, you want to compare differences in velocities `np.diff(param_values[n_refractors:])`.
2) Even if the first one is fixed, `np.isclose` will return `False` if compared values differ in both directions, while in this case we are good if difference between velocities are much higher than `min_velocity_step`. For example, in some cases we deliberately set `min_velocity_step` to 1 and it is OK if `v2-v1` is about 300.

seismicpro/refractor_velocity/refractor_velocity.py
```diff
@@ -35,6 +37,9 @@ class RefractorVelocity:
       This methods allows one to specify initial values of some parameters or bounds for their values or simply provide
       the expected number of refractors,
     * `from_file` - to create a velocity model from parameters stored in a file.
+    * `from_survey` - to automatically fit a near-surface velocity model by offsets and times of first breaks stored in
+      the survey. This method can be initialzed with no initial values, bounds, or expected number of refractors and is
+      helpful for exploratory data analysis of a survey.
```
nikita-klsh, line 42:
```suggestion
    * `from_survey` - esentially the same as `from_first_breaks` except that near-surface velocity model is fitted for the whole survey(represented by one large supergather). 
        Helpful for exploratory data analysis of a survey, e.g. before running `RVF.from_survey()`. 
```

seismicpro/refractor_velocity/refractor_velocity.py
```diff
@@ -35,6 +37,9 @@ class RefractorVelocity:
       This methods allows one to specify initial values of some parameters or bounds for their values or simply provide
       the expected number of refractors,
```
nikita-klsh, line 38:
```suggestion
      the expected number of refractors. Ultimately, none of the paramteres can be provided, then number of refractors heuristically estimated
```

seismicpro/refractor_velocity/utils.py
```diff
@@ -91,3 +92,64 @@ def dump_refractor_velocities(refractor_velocities, path, encoding="UTF-8"):
...
+    separately by bins.
+    """
+    headers = survey.headers[['offset', first_breaks_col]].copy()
+    headers['bins'] = headers['offset'] // reduce_step
+    reduced_headers = headers.groupby(by='bins', sort=False).mean()
+    return reduced_headers['offset'].to_numpy(), reduced_headers[first_breaks_col].to_numpy()
+
+
+def calc_optimal_velocity(offsets, times, init=None, bounds=None, min_velocity_step=400, min_refractor_size=600,
+                          loss="L1", huber_coef=20, min_refractors=1, max_refractors=10):
```
nikita-klsh, line 111: I would not create new function for this, and move this code right to the `from_first_breaks`
There is a lot of estimating logic there, no problem to add this aswell

Moreover, explicitely setting `min_velocity_step=400` `and min_refractor_size=600` there gives user a good hint how your heuristic works

seismicpro/refractor_velocity/refractor_velocity_field.py
```diff
@@ -241,7 +244,11 @@ def from_survey(cls, survey, is_geographic=None, auto_create_interpolator=True,
             If any gather has non-unique pair of coordinates.
         """
         if survey.n_gathers < 1:
-            raise ValueError("Survey is empty")
+            raise ValueError("Survey is empty.")
+        if all(param is None for param in (init, bounds, n_refractors)):
+            offsets, times = reduce_offsets_and_times(survey, first_breaks_col)
+            init = RefractorVelocity.from_first_breaks(offsets, times).params
```
nikita-klsh, line 250: Why not?
```suggestion
            init = RefractorVelocity.from_survey(survey).params
```

seismicpro/refractor_velocity/utils.py
```diff
@@ -91,3 +92,64 @@ def dump_refractor_velocities(refractor_velocities, path, encoding="UTF-8"):
         data[i] = [*rv.coords.names] + [*rv.coords.coords] + list(rv.params.values())
     df = pd.DataFrame(data, columns=columns).convert_dtypes()
     df.to_string(buf=path, float_format=lambda x: f"{x:.2f}", index=False, encoding=encoding)
+
+
+def reduce_offsets_and_times(survey, first_breaks_col=HDR_FIRST_BREAK, reduce_step=20):
```
nikita-klsh, line 97: If we use this function once, do not create a new function and use the code directly

seismicpro/refractor_velocity/utils.py
```diff
@@ -91,3 +92,64 @@ def dump_refractor_velocities(refractor_velocities, path, encoding="UTF-8"):
...
+    """Reduce the offsets and times of first breaks stored in the survey headers.
+
+    Method splits the survey headers by bins with the same size define by `reduce_step` using offsets values and
+    calculate the mean separately by bins. The first breaks times uses same bins and also calculate the mean
+    separately by bins.
+    """
+    headers = survey.headers[['offset', first_breaks_col]].copy()
+    headers['bins'] = headers['offset'] // reduce_step
+    reduced_headers = headers.groupby(by='bins', sort=False).mean()
+    return reduced_headers['offset'].to_numpy(), reduced_headers[first_breaks_col].to_numpy()
```
nikita-klsh, line 107: No need to create intermediate column
```suggestion
    return headers.groupby(headers['offset'] // reduce_step, sort=False).mean().values.T
```

seismicpro/refractor_velocity/utils.py
bestetc: move to `RefractorVelocity.from_survey` method.

seismicpro/refractor_velocity/utils.py
bestetc: I don't like adding this code to the from_first_break method.
1. `from_first_breaks` is already overloaded with code.
2. `calc_optimal_velocity` logic is recursive and looks strange in `from_first_breaks`.

I will move this function to RefractorVelocity as a separate method.

seismicpro/refractor_velocity/utils.py
```diff
@@ -91,3 +92,64 @@ def dump_refractor_velocities(refractor_velocities, path, encoding="UTF-8"):
...
+    #pylint: disable-next=import-outside-toplevel
+    from .refractor_velocity import RefractorVelocity  # avoid circular import
+    rv_base = RefractorVelocity.from_first_breaks(offsets, times, n_refractors=min_refractors, loss=loss,
+                                                  huber_coef=huber_coef)
+    for refractor in range(min_refractors + 1, max_refractors + 1):
+        max_offset = max(offsets.max(), min_refractor_size * refractor)
+        rv_last = RefractorVelocity.from_first_breaks(offsets, times, init, bounds, refractor, max_offset,
+                                                      min_velocity_step, min_refractor_size, loss, huber_coef)
+        n_points, _ = np.histogram(rv_last.offsets, bins=rv_last.piecewise_offsets)
+        if (n_points < 1).any() or (rv_last.fit_result.fun > rv_base.fit_result.fun):
```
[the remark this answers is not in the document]
bestetc: Yes, I try to fix random seed with some big prime number and different launch give the same result now.

Also I try to find defaults parameters again with fixed seed. Grid search give the same result.
Figure 30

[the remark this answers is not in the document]
bestetc: Does we need fix random seed for SGDRegressor for RefractorVelocity or fixed it for some cases?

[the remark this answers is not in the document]
nikita-klsh: Definetely need to fix seed in `calc_optimal_velocity` call for your heuristic
We ca fix it everywhere, so `gather.calculate_refractor_velocity(n_refractors=2)` gives the same results from call to call as well

alexanderkuvaev (changes requested):

seismicpro/refractor_velocity/refractor_velocity.py
```diff
@@ -209,7 +213,7 @@ def from_first_breaks(cls, offsets, times, init=None, bounds=None, n_refractors=
         min_refractor_size = np.ceil(min_refractor_size)
 
         if all(param is None for param in (init, bounds, n_refractors)):
-            raise ValueError("At least one of `init`, `bounds` or `n_refractors` must be defined")
+            init = cls.calc_optimal_velocity(offsets, times, loss=loss, huber_coef=huber_coef).params
```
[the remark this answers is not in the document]
bestetc: Yes, but number of refractors doesn't change.

seismicpro/refractor_velocity/refractor_velocity.py
```diff
@@ -608,6 +664,50 @@ def calculate_loss(cls, scaled_params, offsets, times, max_offset, loss='L1', hu
...
+        min_refractor_size : int, optional, defaults to 650
+            Minimum offset range covered by each refractor.
+        loss : str, optional, defaults to "L1"
+            Loss function to be minimized. Should be one of "MSE", "huber", "L1", "soft_L1", or "cauchy".
+        huber_coef : float, optional, default to 20
+            Coefficient for Huber loss function.
+        min_refractors : int, optional, defaults to 1
+            Minimum number of refractors for the expected velocity model.
+        max_refractors : int, optional, defaults to 10
+            Maximum number of refractors for the expected velocity model.
```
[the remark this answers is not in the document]
bestetc: moved

seismicpro/refractor_velocity/refractor_velocity.py
```diff
@@ -608,6 +664,50 @@ def calculate_loss(cls, scaled_params, offsets, times, max_offset, loss='L1', hu
...
+
+        Returns
+        -------
+        rv : RefractorVelocity
+            A near-surface velocity model with optimal number of refractors.
+        """
+        rv_base = cls.from_first_breaks(offsets, times, n_refractors=min_refractors, loss=loss, huber_coef=huber_coef)
+        for refractor in range(min_refractors + 1, max_refractors + 1):
+            max_offset = max(offsets.max(), min_refractor_size * refractor)
```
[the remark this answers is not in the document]
bestetc: If this  method tries to check hypotesys that velocity model has 3 refractors with size 500, it means that max offset is at least 1500.  Just it.

seismicpro/refractor_velocity/refractor_velocity.py
```diff
@@ -608,6 +664,50 @@ def calculate_loss(cls, scaled_params, offsets, times, max_offset, loss='L1', hu
...
+        Parameters
+        ----------
+        offsets : 1d ndarray
+            Offsets of traces. Measured in meters.
+        times : 1d ndarray
+            Time of first break for each trace. Measured in milliseconds.
+        init : dict, optional
+            Initial values of model parameters.
+        bounds : dict, optional
+            Lower and upper bounds of model parameters.
```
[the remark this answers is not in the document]
bestetc: removed

seismicpro/refractor_velocity/refractor_velocity.py
```diff
@@ -209,7 +213,7 @@ def from_first_breaks(cls, offsets, times, init=None, bounds=None, n_refractors=
         min_refractor_size = np.ceil(min_refractor_size)
 
         if all(param is None for param in (init, bounds, n_refractors)):
```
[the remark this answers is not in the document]
bestetc: Updated. Check that

PR #115, opened 2022-11-16
Return a 1d-array for a single header selection 

alexanderkuvaev:
1. Now the following code
    ```python
    offset = gather["offset"]
    ```
    returns a 1d array instead of 2d. Old behavior can be obtained by always passing a list of headers similar to `pandas` convention:
    ```python
    offset_2d = gather[["offset"]]
    ```

2. Now `__setitem__` of trace containers acts identically to that of a `pd.DataFrame`. Previously the following code worked:
    ```python
    survey[["my_header"]] = np.arange(survey.n_traces)
    ```
    Now, following the `pandas` convention, it raises an error. Two valid options are:
    ```python
    survey["my_header"] = np.arange(survey.n_traces)
    survey[["my_header"]] = np.arange(survey.n_traces).reshape(-1, 1)
    ```

3. Simplifies `IndexPart` instantiation a bit.

### DIFF
--- a/seismicpro/containers.py
+++ b/seismicpro/containers.py
@@ -12,7 +12,7 @@ import numpy as np
 import pandas as pd
 
 from .decorators import batch_method
-from .utils import to_list, get_cols, set_cols, create_indexer, maybe_copy
+from .utils import to_list, get_cols, create_indexer, maybe_copy
 
 
 class SamplesContainer:
@@ -63,7 +63,7 @@ class TraceContainer:
 
         Returns
         -------
-        result : 2d np.ndarray
+        result : np.ndarray
             Headers values.
         """
         return get_cols(self.headers, key)
@@ -78,7 +78,7 @@ class TraceContainer:
         value : np.ndarray
             Headers values to set.
         """
-        return set_cols(self.headers, key, value)
+        self.headers[key] = value
 
     def copy(self, ignore=None):
         """Perform a deepcopy of all attributes of `self` except for those specified in `ignore`, which are kept
--- a/seismicpro/containers.py
+++ b/seismicpro/containers.py
@@ -12,7 +12,7 @@ import numpy as np
 import pandas as pd
 
 from .decorators import batch_method
-from .utils import to_list, get_cols, create_indexer, maybe_copy
+from .utils import to_list, get_cols, set_cols, create_indexer, maybe_copy
 
 
 class SamplesContainer:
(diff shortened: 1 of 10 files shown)


seismicpro/utils/general_utils.py
```diff
@@ -48,25 +49,46 @@ def get_first_defined(*args):
...
-        col_values = df[col] if col in df.columns else df.index.get_level_values(col)
-        res.append(col_values.values)
+    for col in cols:
+        if col in df.columns:
+            col_values = df[col]
+        elif col in df.index.names:
+            col_values = df.index.get_level_values(col)
+        else:
+            raise KeyError(f"Unknown header {col}")
+        res.append(col_values.to_numpy())
```
anton-br, line 78: Consider check if any headers are missed before `for` to raise error for all missed headers, not just for the first one.

something like this:
```suggestion
    unknown_headers = set(cols) - set(df.columns)
    if unknown_headers:
        raise KeyError(f"Unknown headers: {', '.join(unknown_headers)}")
    for col in cols:
        if col in df.columns:
            col_values = df[col]
        elif col in df.index.names:
            col_values = df.index.get_level_values(col)
        res.append(col_values.to_numpy())
```

dpodvyaznikov (approved):

seismicpro/utils/general_utils.py
```diff
@@ -48,25 +49,46 @@ def get_first_defined(*args):
...
+
+    if is_single_col:
+        return res[0]
     return np.column_stack(res)
 
 
+def set_cols(df, cols, values):
+    """Set given `values` to selected columns of the `df` DataFrame."""
+    cols, _ = _process_cols(df, cols)
+    values = pd.DataFrame(values).to_numpy()
```
nikita-klsh, line 88: Do you need this intermediate cast to `DataFrame`?

seismicpro/utils/general_utils.py
alexanderkuvaev: The proposal does not handle `MultiIndex` columns which are used by `IndexPart`. The easiest proper way to check whether columns exist in a `DataFrame` requires iterative `in` check which I'd like to call once.

alexanderkuvaev:
Update: now `__setitem__` of trace containers acts identically to that of a `pd.DataFrame`. Previously the following code worked.
```python
survey[["my_header"]] = np.arange(survey.n_traces)
```
Now, following the `pandas` convention, it raises an error. Two valid options are:
```python
survey["my_header"] = np.arange(survey.n_traces)
survey[["my_header"]] = np.arange(survey.n_traces).reshape(-1, 1)
```
This is a breaking change, requesting review again.

seismicpro/utils/general_utils.py
alexanderkuvaev: See my comment below

seismicpro/containers.py
```diff
@@ -69,7 +66,7 @@ def __getitem__(self, key):
         result : 2d np.ndarray
```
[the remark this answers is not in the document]
alexanderkuvaev: Fixed

anton-br (approved):

nikita-klsh (approved):

PR #117, opened 2022-11-17
Fix refine

alexanderkuvaev:
* Change `_get_refined_values` logic:
  * Now only ignored values are interpolated,
  * Now `create_interpolator` passes all interpolator `kwargs` for refinement.
* Speed up `RefractorVelocityField.from_survey` by ~2x for surveys with a large number of small (~100 traces) gathers (e.g. indexed by (`INLINE_3D`, `CROSSLINE_3D`))
* Switch from `NearestNeighbors` to `KDTree` in a couple of places to speed things up

### DIFF
--- a/seismicpro/refractor_velocity/refractor_velocity_field.py
+++ b/seismicpro/refractor_velocity/refractor_velocity_field.py
@@ -153,14 +153,15 @@ class RefractorVelocityField(SpatialField):
         return msg
 
     @staticmethod
-    def _fit_refractor_velocities(rv_kwargs_list):
+    def _fit_refractor_velocities(rv_kwargs_list, common_kwargs):
         """Fit a separate near-surface velocity model by offsets and times of first breaks for each set of parameters
         defined in `rv_kwargs_list`. This is a helper function and is defined as a `staticmethod` only to be picklable
         so that it can be passed to `ProcessPoolExecutor.submit`."""
-        return [RefractorVelocity.from_first_breaks(**rv_kwargs) for rv_kwargs in rv_kwargs_list]
+        return [RefractorVelocity.from_first_breaks(**rv_kwargs, **common_kwargs) for rv_kwargs in rv_kwargs_list]
 
     @classmethod
-    def _fit_refractor_velocities_parallel(cls, rv_kwargs_list, chunk_size=250, n_workers=None, bar=True, desc=None):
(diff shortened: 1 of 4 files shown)


seismicpro/refractor_velocity/refractor_velocity_field.py
```diff
@@ -240,28 +243,35 @@ def from_survey(cls, survey, is_geographic=None, auto_create_interpolator=True,
...
-        if survey.n_gathers < 1:
+        if survey.is_empty:
             raise ValueError("Survey is empty")
+        if survey.headers.index.is_unique:
+            raise ValueError("Survey must be indexed by gathers, not individual traces")
+
+        # Extract required headers for each gather in the survey
         coords_cols = get_coords_cols(survey.indexed_by)
         survey_headers = survey[("offset", first_breaks_col) + coords_cols]
+        gather_change_ix = np.where(~survey.headers.index.duplicated())[0][1:]
```
a-arefina, line 254: maybe add explicitly `keep='first'` to indicate that you are actually marking the first occurrence of each index?

nikita-klsh, line 254: We already have  this precomputed: `survey._indexer.index_to_pos`

seismicpro/refractor_velocity/refractor_velocity_field.py
```diff
@@ -240,28 +243,35 @@ def from_survey(cls, survey, is_geographic=None, auto_create_interpolator=True,
             If the survey is empty.
             If any gather has non-unique pair of coordinates.
         """
-        if survey.n_gathers < 1:
+        if survey.is_empty:
             raise ValueError("Survey is empty")
+        if survey.headers.index.is_unique:
```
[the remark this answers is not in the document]
alexanderkuvaev: Because it makes little sence to fit refractor velocity by a single point. Added to docs.

seismicpro/refractor_velocity/refractor_velocity_field.py
alexanderkuvaev: Added

alexanderkuvaev: We have, but it's stored as a dict of ranges and additional transformation of this structure into an array used here is required. Moreover, `index_to_pos` is created only for one subclass of base indexer. Currently it is used anyways because of the `if survey.headers.index.is_unique` check above, but removing it some time will make us fixing it. Kept it as is for now.

nikita-klsh (approved):

PR #120, opened 2022-12-08
Sort gather by multiple headers

nikita-klsh:
`gather.sort()` now can work with multiple headers, e.g. `gather.sort(['INLINE_3D', 'CROSSLINE_3D'])`

The `gather.sort_by` attribute is latter is used in `gather.plot()` to infer axis ticks. 
In case `gather` is sorted by two headers, add minor ticks to handle minor header, e.g.

`gather.sort(['SourceWaterDepath', 'GroupWaterDepth']).plot()`
image

### DIFF
--- a/seismicpro/gather/gather.py
+++ b/seismicpro/gather/gather.py
@@ -189,11 +189,13 @@ class Gather(TraceContainer, SamplesContainer):
         # The two-element `indices` tuple describes indices of traces and samples to be obtained respectively
         new_self.headers = self.headers.iloc[indices[0]]
         new_self.samples = self.samples[indices[1]]
-
-        # Check that `sort_by` still represents the actual trace sorting as it might be changed during getitem.
-        if new_self.sort_by is not None:
-            if (np.lexsort(new_self.headers[to_list(new_self.sort_by)].values.T[::-1]) != np.arange(len(data))).any():
+        
+        # Check that `sort_by` does not represent the actual trace sorting as it might be changed during getitem.
+        if isinstance(indices[0], slice):
+            if indices[0].step is not None and indices[0].step < 0: # Slice with negative step breaks sorting
                 new_self.sort_by = None
+        elif (np.diff(indices[0]) < 0).any(): # Decreasing sequence of indices breaks sorting
+            new_self.sort_by = None
         return new_self
 
     def __str__(self):
@@ -915,10 +917,13 @@ class Gather(TraceContainer, SamplesContainer):
         self : Gather
             Gather sorted by given headers. Sets `sort_by` attribute to `by`.
         """
-        if self.sort_by == by:
+        by = to_list(by)
+        sort_by = to_list(self.sort_by)
+        if len(sort_by) >= len(by) and all(i == j for i, j in zip(sort_by, by)):
             return self
-        order = np.lexsort(self[to_list(by)].T[::-1])
-        self.sort_by = by
+
+        order = np.lexsort(self[by[::-1]].T)
+        self.sort_by = by[0] if len(by) == 1 else by
         self.data = self.data[order]
         self.headers = self.headers.iloc[order]
         return self
@@ -1422,8 +1427,9 @@ class Gather(TraceContainer, SamplesContainer):
         if top_header is not None:
             top_ax = self._plot_top_subplot(ax=ax, divider=divider, header_values=self[top_header], y_ticker=y_ticker)
 
-        # Set axis ticks
+        # Set axis ticks. X axis can be represented by more than 2 headers, we plot ticks for no more than 2
         x_tick_src = x_tick_src or self.sort_by or "index"
+        x_tick_src = to_list(x_tick_src)[:2]
         self._set_ticks(ax, axis="x", tick_src=x_tick_src, ticker=x_ticker)
         self._set_ticks(ax, axis="y", tick_src=y_tick_src, ticker=y_ticker)
 
@@ -1499,9 +1505,12 @@ class Gather(TraceContainer, SamplesContainer):
 
     def _get_x_ticks(self, axis_label):
         """Get tick labels for x-axis: either any gather header or ordinal numbers of traces in the gather."""
-        if axis_label == "index":
-            return np.arange(self.n_traces)
-        return self[axis_label]
+        if axis_label[0] == "index":
+            return np.arange(self.n_traces), None
+        elif len(axis_label) == 1:
+            return self[axis_label[0]], None
+        else:
+            return self[axis_label].T
 
     def _get_y_ticks(self, axis_label):
         """Get tick labels for y-axis: either time samples or ordinal numbers of samples in the gather."""
@@ -1515,12 +1524,12 @@ class Gather(TraceContainer, SamplesContainer):
         """Set ticks, their labels and an axis label for a given axis."""
         # Get tick_labels depending on axis and its label
         if axis == "x":
-            tick_labels = self._get_x_ticks(tick_src)
+            major_labels, minor_labels = self._get_x_ticks(tick_src)
         elif axis == "y":
-            tick_labels = self._get_y_ticks(tick_src)
+            major_labels, minor_labels = self._get_y_ticks(tick_src), None
         else:
             raise ValueError(f"Unknown axis {axis}")
-        set_ticks(ax, axis, tick_labels=tick_labels, **{"label": tick_src, **ticker})
+        set_ticks(ax, axis, major_labels=major_labels, minor_labels=minor_labels, **{"label": tick_src, **ticker})
 
     def plot_nmo_correction(self, min_vel=1500, max_vel=6000, figsize=(6, 4.5), show_grid=True, **kwargs):
         """Perform interactive NMO correction of the gather with selected constant velocity.
--- a/seismicpro/gather/gather.py
+++ b/seismicpro/gather/gather.py
@@ -1516,8 +1516,8 @@ class Gather(TraceContainer, SamplesContainer):
         if tick_src == "time":
             major_labels =  self.samples
         if tick_src == "samples":
-            major_labels = np.arange(self.n_samples
-)
+            major_labels = np.arange(self.n_samples)
+
         set_ticks(ax, 'y', major_labels=major_labels, **{"label": tick_src, **ticker})
 
     def plot_nmo_correction(self, min_vel=1500, max_vel=6000, figsize=(6, 4.5), show_grid=True, **kwargs):
--- a/seismicpro/gather/gather.py
+++ b/seismicpro/gather/gather.py
@@ -1520,7 +1520,7 @@ class Gather(TraceContainer, SamplesContainer):
         UNITS = {  # pylint: disable=invalid-name
             "offset": ", m",
         }
-        
+
         tick_src = [ix_tick_src + UNITS.get(ix_tick_src, '') for ix_tick_src in tick_src]
 
         set_ticks(ax, 'x', major_labels=major_labels, minor_labels=minor_labels, **{"label": tick_src, **ticker})
--- a/seismicpro/gather/gather.py
+++ b/seismicpro/gather/gather.py
@@ -191,7 +191,7 @@ class Gather(TraceContainer, SamplesContainer):
         new_self.samples = self.samples[indices[1]]
 
         # Check that `sort_by` still represents the actual trace sorting as it might be changed during getitem.
-        if new_self.sort_by is not None: 
+        if new_self.sort_by is not None:
             if (np.lexsort(new_self.headers[to_list(new_self.sort_by)].values.T[::-1]) != np.arange(len(data))).any():
                 new_self.sort_by = None
         return new_self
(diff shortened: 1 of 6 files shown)


seismicpro/gather/gather.py
```diff
@@ -902,30 +903,21 @@ def apply_nmo(self, stacking_velocity):
...
-        ValueError
-            If `by` column was not loaded in `headers`.
+            Gather sorted by given headers. Sets `sort_by` attribute to `by`.
         """
-        if not isinstance(by, str):
-            raise TypeError(f'`by` should be str, not {type(by)}')
         if self.sort_by == by:
             return self
-        order = np.argsort(self[by], kind='stable')
+        order = np.lexsort(self[to_list(by)].T[::-1])
```
alexanderkuvaev, line 920:
```suggestion
        order = np.lexsort(self[to_list(by)[::-1]].T)
```
For me such order is more clear, but up to you

seismicpro/gather/gather.py
```diff
@@ -191,8 +191,9 @@ def __getitem__(self, key):
         new_self.samples = self.samples[indices[1]]
 
         # Check that `sort_by` still represents the actual trace sorting as it might be changed during getitem.
-        if new_self.sort_by is not None and not new_self.headers[new_self.sort_by].is_monotonic_increasing:
-            new_self.sort_by = None
+        if new_self.sort_by is not None:
+            if (np.lexsort(new_self.headers[to_list(new_self.sort_by)].values.T[::-1]) != np.arange(len(data))).any():
```
alexanderkuvaev, line 195: `__getitem__` was supposed to be as fast as possible, but here a lexsort over all headers is called. Check if we can simply this by setting `sort_by` to `None` if `indices[0]` is not monotonically increasing.

seismicpro/gather/gather.py
```diff
@@ -902,30 +903,21 @@ def apply_nmo(self, stacking_velocity):
...
-        ------
-        TypeError
-            If `by` is not str.
-        ValueError
-            If `by` column was not loaded in `headers`.
+            Gather sorted by given headers. Sets `sort_by` attribute to `by`.
         """
-        if not isinstance(by, str):
-            raise TypeError(f'`by` should be str, not {type(by)}')
         if self.sort_by == by:
```
alexanderkuvaev, line 918: Now you should do nothing if `self.sort_by` starts with `by`

seismicpro/utils/plot_utils.py
```diff
@@ -124,20 +126,33 @@ def set_ticks(ax, axis, label='', tick_labels=None, num=None, step_ticks=None, s
         "Time": " (ms)",
         "Offset": " (m)",
     }
-    label = label[0].upper() + label[1:]
```
alexanderkuvaev, line 127: We still have an old bug: default value for label is an empty string and here we take the first element of it, which fails. Please fix it in your code below

seismicpro/utils/plot_utils.py
```diff
@@ -124,20 +126,33 @@ def set_ticks(ax, axis, label='', tick_labels=None, num=None, step_ticks=None, s
...
-    label += UNITS.get(label, "")
 
-    locator, formatter = _process_ticks(labels=tick_labels, num=num, step_ticks=step_ticks,
+    label = to_list(label)
+    axis_label = '\n'.join([ix_label[0].upper() + ix_label[1:] + UNITS.get(ix_label, '') for ix_label in label])
+
+    major_labels, minor_labels = None, None
+    if len(label) > 1:
```
alexanderkuvaev, line 134: Please add a check that `len(labels)` equals `tick_labels.shape[1]` if it's 2d

alexanderkuvaev (changes requested):

seismicpro/gather/gather.py
```diff
@@ -1523,12 +1524,12 @@ def _set_ticks(self, ax, axis, tick_src, ticker):
         """Set ticks, their labels and an axis label for a given axis."""
         # Get tick_labels depending on axis and its label
         if axis == "x":
-            tick_labels = self._get_x_ticks(tick_src)
+            major_labels, minor_labels = self._get_x_ticks(tick_src)
         elif axis == "y":
-            tick_labels = self._get_y_ticks(tick_src)
+            major_labels, minor_labels = self._get_y_ticks(tick_src), None
```
alexanderkuvaev, line 1529: `_get_x_ticks` and `_get_y_ticks` are supposed to be alike, but the first one returns both major and minor tick labels, while the latter returns only mahor ticks. I'd split `_set_ticks` method into two: `_set_x_ticks` and `_set_y_ticks` which incorporate  the code of `_get_x_ticks` and `_get_y_ticks` respectively. This will make the code cleaner and remove this strange `if`.

alexanderkuvaev (changes requested):

seismicpro/utils/plot_utils.py
```diff
@@ -124,20 +130,26 @@ def set_ticks(ax, axis, label='', tick_labels=None, num=None, step_ticks=None, s
         "Time": " (ms)",
         "Offset": " (m)",
     }
-    label = label[0].upper() + label[1:]
-    label += UNITS.get(label, "")
 
-    locator, formatter = _process_ticks(labels=tick_labels, num=num, step_ticks=step_ticks,
+    axis_label = '\n'.join([UNITS.get(ix_label, '') for ix_label in to_list(label)])
```
anton-br, line 134: this won't work:(
```suggestion
    axis_label = '\n'.join([ix_label + UNITS.get(ix_label, '') for ix_label in to_list(label)])
```

anton-br (approved):
Anyway, consider expand tests for gahter.sort by adding single parameter with multiple headers.

I suggest add smth like this in test/gather_test.py:
```
@pytest.mark.parametrize('by', ('offset', ['offset', 'FieldRecord']))
def test_gather_sort(gather, by):
    """test_gather_sort"""
    gather.sort(by=by)
```

seismicpro/gather/gather.py
```diff
@@ -190,9 +190,13 @@ def __getitem__(self, key):
         new_self.headers = self.headers.iloc[indices[0]]
         new_self.samples = self.samples[indices[1]]
 
-        # Check that `sort_by` still represents the actual trace sorting as it might be changed during getitem.
-        if new_self.sort_by is not None and not new_self.headers[new_self.sort_by].is_monotonic_increasing:
-            new_self.sort_by = None
+        # If the gather was sorted, verify that getitem does not break sorting
+        if new_self.sort_by is not None:
```
alexanderkuvaev, line 194: You don't need the outer `if`

seismicpro/utils/plot_utils.py
```diff
@@ -121,23 +127,32 @@ def set_ticks(ax, axis, label='', tick_labels=None, num=None, step_ticks=None, s
     """
     # Format axis label
     UNITS = {  # pylint: disable=invalid-name
```
alexanderkuvaev, line 129: This dict becomes extremely complicated: now keys like "Relative velocity margin" are present here. Can we remove this units logic at all and move it in the outer calling function?

seismicpro/gather/gather.py
nikita-klsh: Tho it can be removed, but then unnecessary computations would be done in case gather was not sorted.

I would stick with master logic where it could be remove aswell, but it presents.

seismicpro/gather/gather.py
```diff
@@ -1329,21 +1325,21 @@ def plot(self, mode="seismogram", *, title=None, x_ticker=None, y_ticker=None, a
...
+        set_ticks(ax, "y", **{"label": "Counts", **y_ticker})
 
         ax.grid(grid)
         if log:
             ax.set_yscale("log")
         ax.set_title(**{'label': None, **title})
 
     # pylint: disable=too-many-arguments
-    def _plot_seismogram(self, ax, title, x_ticker, y_ticker, x_tick_src=None, y_tick_src='time', colorbar=False,
+    def _plot_seismogram(self, ax, title, x_ticker, y_ticker, x_tick_src=None, y_tick_src='Time', colorbar=False,
```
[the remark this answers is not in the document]
nikita-klsh: Fixed, thx

nikita-klsh:
> Anyway, consider expand tests for gahter.sort by adding single parameter with multiple headers.
> 
> I suggest add smth like this in test/gather_test.py:
> 
> ```
> @pytest.mark.parametrize('by', ('offset', ['offset', 'FieldRecord']))
> def test_gather_sort(gather, by):
>     """test_gather_sort"""
>     gather.sort(by=by)
> ```
Added

seismicpro/utils/plot_utils.py
nikita-klsh: If only the presence of non physical unit among physical ones looks redundant, we can remove only him.

The pros of having such dict: all the units logic happens in one place. For example, if once we would have foots for distance, feels like it would be easier to catch up.

seismicpro/gather/gather.py
```diff
@@ -902,31 +906,24 @@ def apply_nmo(self, stacking_velocity):
...
-        TypeError
-            If `by` is not str.
-        ValueError
-            If `by` column was not loaded in `headers`.
+            Gather sorted by given headers. Sets `sort_by` attribute to `by`.
         """
-        if not isinstance(by, str):
-            raise TypeError(f'`by` should be str, not {type(by)}')
-        if self.sort_by == by:
+        by = to_list(by)
```
a-arefina, line 921: Maybe check add some checks on `by`? Ex., `len(by) > 0` & each element is `str`

seismicpro/gather/gather.py
```diff
@@ -1329,21 +1325,21 @@ def plot(self, mode="seismogram", *, title=None, x_ticker=None, y_ticker=None, a
         plotters_dict[mode](ax, title=title, x_ticker=x_ticker, y_ticker=y_ticker, **kwargs)
         return self
 
-    def _plot_histogram(self, ax, title, x_ticker, y_ticker, x_tick_src="amplitude", bins=None,
+    def _plot_histogram(self, ax, title, x_ticker, y_ticker, x_tick_src="Amplitude", bins=None,
                         log=False, grid=True, **kwargs):
         """Plot histogram of the data specified by x_tick_src."""
-        data = self.data.ravel() if x_tick_src == "amplitude" else self[x_tick_src]
+        data = self.data.ravel() if x_tick_src == "Amplitude" else self[x_tick_src]
```
a-arefina, line 1331: what about allowing both lower case and title case by adding `x_tick_src = x_tick_src.title()` ?

seismicpro/gather/gather.py
alexanderkuvaev: Fine, thx

seismicpro/utils/plot_utils.py
alexanderkuvaev: I don't like that a function that should only set ticks also alters the passed parameters (`label` in this case). Such an approach would make even harder to handle other units like distances in feet because only outer object such as a `Gather` knows about it.

I'd do the following:
1. Remove this extra units processing logic and set `label` as it was passed,
2. Restore `Semblance`/`RefractorVelocity`/... plots back to `master` version which processed units on their own,
3. Move units processing logic inside `_set_x_ticks` and `_set_y_ticks` of a `Gather` class

This will also allow to un-capitalize default arguments of all `Gather._plot*` methods as suggested by @a-arefina

alexanderkuvaev (changes requested):

seismicpro/gather/gather.py
```diff
@@ -1505,30 +1505,35 @@ def _plot_top_subplot(self, ax, divider, header_values, y_ticker, **kwargs):
...
+
+        tick_src = [ix_tick_src + UNITS.get(ix_tick_src, '') for ix_tick_src in tick_src]
+
+        set_ticks(ax, 'x', major_labels=major_labels, minor_labels=minor_labels, **{"label": tick_src, **ticker})
+
+    def _set_y_ticks(self, ax, tick_src, ticker):
+        """Infer and set ticks for y axis. """
+        tick_src = tick_src.title()
+        if tick_src == "Time, ms":
+            major_labels =  self.samples
```
alexanderkuvaev, line 1532: `tick_src` is `"time"` here, `", ms"` is added later. Moreover, `.title` should have capitalized it to `"Time, Ms"`. And remove extra space after =.

alexanderkuvaev (changes requested):

seismicpro/utils/plot_utils.py
```diff
@@ -119,25 +125,25 @@ def set_ticks(ax, axis, label='', tick_labels=None, num=None, step_ticks=None, s
...
-    # Format axis label
-    UNITS = {  # pylint: disable=invalid-name
-        "Time": " (ms)",
-        "Offset": " (m)",
-    }
-    label = label[0].upper() + label[1:]
-    label += UNITS.get(label, "")
-
-    locator, formatter = _process_ticks(labels=tick_labels, num=num, step_ticks=step_ticks,
+    axis_label = '\n'.join(to_list(label))
```
alexanderkuvaev, line 128: Let's move this line to the only place it is currently used: `Gather._set_x_ticks`

alexanderkuvaev (changes requested):

alexanderkuvaev (approved):

PR #123, opened 2023-01-18
Fields bugfixes

alexanderkuvaev:
Fixed several small bugs appeared during processing of last surveys:
* Now `RefractorVelocity.__getattr__` works as expected: previously `rv.t0` failed,
* Added tolerances in `RefractorVelocity._validate_params` checks. Previously they may have failed to fit a velocity model for noisy first breaks. Was also mentioned in https://github.com/gazprom-neft/SeismicPro/pull/107, but not completely fixed,
* Fixed a rare bug when IDW interpolation within a given radius resulted in `nan` values due to inaccuracies in floating-point arithmetics,
* Fixed processing of `coords_cols` during a field update: e.g. previously `RefractorVelocityField.from_survey(survey, n_refractors=1, is_geographic=True)` produced wrong results if `survey` was indexed by line-based coordinates.

### DIFF
--- a/seismicpro/field.py
+++ b/seismicpro/field.py
@@ -75,8 +75,11 @@ class Field:
         Whether coordinate system of the field is geographic. `None` for an empty field if was not specified during
         instantiation.
     coords_cols : tuple with 2 elements or None
-        Names of SEG-Y trace headers representing coordinates of items in the field. `None` if names of coordinates are
-        mixed.
+        Names of SEG-Y trace headers representing coordinates of items in the field if names are the same among all the
+        items and match the geographic system of the field. ("X", "Y") for a field in geographic coordinate system if
+        names of coordinates of its items are either mixed or line-based. ("INLINE_3D", "CROSSLINE_3D") for a field in
+        line-based coordinate system if names of coordinates of its items are either mixed or geographic. `None` for an
+        empty field.
     interpolator : SpatialInterpolator or None
         Field data interpolator.
     is_dirty_interpolator : bool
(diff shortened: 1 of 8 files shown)


seismicpro/field.py
```diff
@@ -299,10 +301,14 @@ def update(self, items):
...
             is_geographic = items[0].coords.is_geographic
 
+        items_match_is_geographic = all(item.coords.is_geographic is is_geographic for item in items)
         coords_cols_set = {item.coords.names for item in items}
-        coords_cols = coords_cols_set.pop() if len(coords_cols_set) == 1 else None
-        if not self.is_empty and coords_cols != self.coords_cols:
-            coords_cols = None
+        if not self.is_empty:
+            coords_cols_set.add(self.coords_cols)
+        if len(coords_cols_set) == 1 and items_match_is_geographic:
```
AnnaAltynova, line 308: maybe we could set `self.is_geographic=True` here in case it was `None`?

seismicpro/field.py
```diff
@@ -75,8 +75,11 @@ class Field:
         Whether coordinate system of the field is geographic. `None` for an empty field if was not specified during
         instantiation.
     coords_cols : tuple with 2 elements or None
-        Names of SEG-Y trace headers representing coordinates of items in the field. `None` if names of coordinates are
-        mixed.
+        Names of SEG-Y trace headers representing coordinates of items in the field if names are the same among all the
+        items and match the geographic system of the field. ("X", "Y") for a field in geographic coordinate system if
+        names of coordinates of its items are either mixed or line-based. ("INLINE_3D", "CROSSLINE_3D") for a field in
```
AnnaAltynova, line 80: I've got a little confused: isn't `("X", "Y")` for geographic and mixed items and `("INLINE_3D", "CROSSLINE_3D")` for line-based and mixed?

seismicpro/field.py
alexanderkuvaev: It's set below

seismicpro/field.py
alexanderkuvaev: Discussed in person

AnnaAltynova (approved):

nikita-klsh (approved):

PR #127, opened 2023-01-31
Handle uphole correction

alexanderkuvaev:
- `Survey` now has an `is_uphole` property, which is calculated by values of `SourceUpholeTime` and `SourceDepth` headers if they are loaded. Consistency of these headers is now checked in `validate_headers`.
- `Gather.calculate_refractor_velocity` and `RefractorVelocityField.from_survey` now accept a `correct_uphole` flag which specifies whether to perform uphole correction by adding values of `SourceUpholeTime` header to times of first breaks or not. This emulates the case when sources are located on the surface. By default, correction is performed if `SourceUpholeTime` header is loaded.
- `RefractorVelocity` now has an `is_uphole_corrected` attribute which is set on construction. `RefractorVelocityField` now has an `is_uphole_corrected` property which is inferred by its items.
- `Gather.apply_lmo` and `FirstBreaksOutliers` metric now properly handle uphole data via their `correct_uphole` flag.
- Upholes are by design not handled in `Muter`, constructed from `RefractorVelocity`: although it seems useful for shot gathers, muting boundary becomes noisy for CDP gathers, which contain traces from multiple shots with varying uphole times. Moreover, similar behavior may be obtained by increasing `delay` on muter construction.

### DIFF
--- a/seismicpro/gather/gather.py
+++ b/seismicpro/gather/gather.py
@@ -884,8 +884,8 @@ class Gather(TraceContainer, SamplesContainer):
             Headers columns which will be LMO-corrected inplace.
         correct_uphole : bool, optional
             Whether to perform uphole correction by adding values of "SourceUpholeTime" header to estimated delay
-            times. If not given, correction is performed if "SourceUpholeTime" header is loaded and given
-            `refractor_velocity` was also uphole corrected.
+            times. If enabled, centers first breaks around `delay` for uphole surveys. If not given, correction is
+            performed if "SourceUpholeTime" header is loaded and given `refractor_velocity` was also uphole corrected.
 
         Returns
         -------
--- a/seismicpro/gather/gather.py
+++ b/seismicpro/gather/gather.py
@@ -865,7 +865,7 @@ class Gather(TraceContainer, SamplesContainer):
     #------------------------------------------------------------------------#
 
     @batch_method(target="threads", args_to_unpack="refractor_velocity")
-    def apply_lmo(self, refractor_velocity, delay=100, fill_value=np.nan, event_headers=None):
+    def apply_lmo(self, refractor_velocity, delay=100, fill_value=np.nan, event_headers=None, correct_uphole=None):
         """Perform gather linear moveout correction using the given near-surface velocity model.
 
         Parameters
@@ -882,6 +882,10 @@ class Gather(TraceContainer, SamplesContainer):
             Value used to fill the amplitudes outside the gather bounds after moveout.
         event_headers : str, list, or None, optional, defaults to None
             Headers columns which will be LMO-corrected inplace.
+        correct_uphole : bool, optional
+            Whether to perform uphole correction by adding values of "SourceUpholeTime" header to estimated delay
+            times. If not given, correction is performed if "SourceUpholeTime" header is loaded and given
+            `refractor_velocity` was also uphole corrected.
 
         Returns
         -------
@@ -902,6 +906,10 @@ class Gather(TraceContainer, SamplesContainer):
                              "type")
 
         trace_delays = delay - refractor_velocity(self.offsets)
+        if correct_uphole:
+            correct_uphole = "SourceUpholeTime" in self.available_headers and refractor_velocity.is_uphole_corrected
+        if correct_uphole:
+            trace_delays += self["SourceUpholeTime"]
         trace_delays_samples = times_to_indices(trace_delays, self.samples, round=True).astype(int)
         self.data = correction.apply_lmo(self.data, trace_delays_samples, fill_value)
         if event_headers is not None:
--- a/seismicpro/gather/gather.py
+++ b/seismicpro/gather/gather.py
@@ -906,7 +906,7 @@ class Gather(TraceContainer, SamplesContainer):
                              "type")
 
         trace_delays = delay - refractor_velocity(self.offsets)
-        if correct_uphole:
+        if correct_uphole is None:
             correct_uphole = "SourceUpholeTime" in self.available_headers and refractor_velocity.is_uphole_corrected
         if correct_uphole:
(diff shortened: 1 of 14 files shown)


seismicpro/gather/metrics.py
```diff
@@ -20,7 +22,7 @@ class FirstBreaksOutliers(PipelineMetric):
     args_to_unpack = ("gather", "refractor_velocity")
 
     @staticmethod
-    def calc(gather, refractor_velocity, first_breaks_col=HDR_FIRST_BREAK, threshold_times=50):
+    def calc(gather, refractor_velocity, first_breaks_col=HDR_FIRST_BREAK, threshold_times=50, correct_uphole=None):
```
AnnaAltynova, line 25: would it be convenient if we allowed `refractor_velocity` to be field instance here? it's more handy in pipelines that way

seismicpro/survey/survey.py
```diff
@@ -311,6 +311,17 @@ def dead_traces_marked(self):
         """bool: `mark_dead_traces` called."""
         return self.n_dead_traces is not None
 
+    @property
+    def is_uphole(self):
```
AnnaAltynova, line 315: do we use it anywhere except info?

seismicpro/survey/survey.py
```diff
@@ -355,6 +366,7 @@ def __str__(self):
         Sample rate:               {self.sample_rate} ms
         Times range:               [{min(self.samples)} ms, {max(self.samples)} ms]
         Offsets range:             {offset_range}
+        Is uphole:                 {"Unknown" if self.is_uphole is None else self.is_uphole}
```
AnnaAltynova, line 369: maybe it would be useful to print uphole range?

seismicpro/survey/survey.py
alexanderkuvaev: Currently no, but it will be later used in the `statics` branch

seismicpro/survey/survey.py
alexanderkuvaev: The range itself does not make a lot of sence, but you can visualize its distribution over the field using `construct_attribute_map`

seismicpro/gather/metrics.py
alexanderkuvaev: Fixed, thx

AnnaAltynova (approved):

seismicpro/refractor_velocity/utils.py
```diff
@@ -85,9 +87,8 @@ def dump_refractor_velocities(refractor_velocities, path, encoding="UTF-8"):
         File encoding.
     """
     rv_list = to_list(refractor_velocities)
-    columns = ['name_x', 'name_y', 'x', 'y'] + list(rv_list[0].params.keys())
-    data = np.empty((len(rv_list), len(columns)), dtype=object)
-    for i, rv in enumerate(rv_list):
-        data[i] = [*rv.coords.names] + [*rv.coords.coords] + list(rv.params.values())
-    df = pd.DataFrame(data, columns=columns).convert_dtypes()
```
nikita-klsh, line 92: Looks like this code is easily expands to your case, why refactor?

seismicpro/survey/survey.py
```diff
@@ -311,6 +311,17 @@ def dead_traces_marked(self):
...
 
+    @property
+    def is_uphole(self):
+        """bool or None: Whether the survey is uphole. `None` if uphole-related headers are not loaded."""
+        has_uphole_times = "SourceUpholeTime" in self.available_headers
+        has_uphole_depths = "SourceDepth" in self.available_headers
+        has_positive_uphole_times = has_uphole_times and (self["SourceUpholeTime"] > 0).any()
+        has_positive_uphole_depths = has_uphole_depths and (self["SourceDepth"] > 0).any()
+        if not has_uphole_times and not has_uphole_depths:
+            return None
```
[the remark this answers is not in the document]
alexanderkuvaev: I'd like to keep all returns close to each other and now it does not affect performance

seismicpro/refractor_velocity/utils.py
alexanderkuvaev: It does not: `convert_dtypes` used here and in the load method acts differently depending on the subset of {`None`, `True`, `False`} values of `is_uphole_corrected` attribute of saved/loaded items: sometimes it becomes `str`, sometimes `bool`. This also can not be unified using `convert_dtypes` arguments.

seismicpro/gather/gather.py
```diff
@@ -740,9 +744,14 @@ def calculate_refractor_velocity(self, init=None, bounds=None, n_refractors=None
         rv : RefractorVelocity
             Constructed near-surface velocity model.
         """
-        return RefractorVelocity.from_first_breaks(self.offsets, self[first_breaks_col], init, bounds, n_refractors,
-                                                   max_offset, min_velocity_step, min_refractor_size, loss, huber_coef,
-                                                   tol, coords=self.coords, **kwargs)
+        times = self[first_breaks_col]
+        if correct_uphole is None:
+            correct_uphole = "SourceUpholeTime" in self.available_headers
```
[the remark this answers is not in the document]
nikita-klsh: What if we introduce survey properties `is_uphole_times` and `is_uphole_depths`
so `is_uphole = is_uphole_times or is_uphole_depths`

And here refer to `is_uphole_times`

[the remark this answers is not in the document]
alexanderkuvaev: This line should also be executed for vibroseis surveys if `SourceUpholeTime` header is loaded, but equals to zero so that the resulting velocity model has `is_uphole_corrected` flag set to `True`. I can make `has_uphole_times` property but not sure if we really need it.

nikita-klsh (approved):

